PR #2, opened 2022-02-17
SQL-732: Adding initial connector files. Updating evergreen build file

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-732

Initial connector files are from [connector-plugin-sdk](https://github.com/tableau/connector-plugin-sdk/tree/master/samples/plugins/postgres_jdbc) with minor changes to allow connection to ADL.  Further customizations will be covered in [SQL-733](https://jira.mongodb.org/browse/SQL-733).

Evergreen task `upload packager log file` is currently failing as the `aws_key` and `aws_secret` is still to be provided by Evergreen team.  I will hold off on merging until the Evergreen build is successful.


evergreen.yml
```diff
@@ -41,6 +41,44 @@ functions:
...
+      params:
+        shell: bash
+        working_dir: mongo-tableau-connector
+        script: |
+          ${prepare_shell}
+          source ./.venv/bin/activate
+          # Running packager validation step
+          ( cd connector-plugin-sdk/connector-packager && python3 -m connector_packager.package \
+          --validate-only --verbose ../../connector/ --log ../../logs 2>&1 | tee /dev/stdout \
+          | grep "Validation succeeded." >/dev/null)
```
bucaojit, line 80: Added these grep commands as the python call was exiting with `0` on success and failure.

connector/connectionProperties.js
```diff
@@ -0,0 +1,14 @@
+(function propertiesbuilder(attr) {
+    var props = {};
```
mattChiaravalloti, line 2: The ADL docs indicate the `compressors` and `authSource` properties should also be included ([see here](https://docs.mongodb.com/datalake/tutorial/jdbc-driver/#configure-the-driver-for-your-tableau)). I've yet to test _without_ those, so maybe we should check before including them. However, if they are necessary then we'll need to include them here.

The JDBC driver takes those properties (and any others in the props file) and adds them to the connection string.

mattChiaravalloti (approved):
Looks good other than my 1 comment. That work could also be pushed to SQL-733, though! lmk what you think.

connector/connectionProperties.js
bucaojit: Included `compressors` and `authSource` and confirmed Tableau connects and can query db metadata and data.

mattChiaravalloti: Maybe this project would be a good time to audit how necessary these two properties are for ADL. We could ask Tom or Ben if they have any idea on this.

bucaojit: I checked with @henretty and he's taking a look at these properties.

connector/connectionBuilder.js
```diff
@@ -0,0 +1,5 @@
+(function dsbuilder(attr) {
+    var urlBuilder = "jdbc:mongodb://" + attr[connectionHelper.attributeServer] + ":" + attr[connectionHelper.attributePort] + "/admin?";
```
nbagnard, line 2: I am not sure where/how this connectionBuilder is used, so maybe it's just because I missed something here.
I thought that the auth source was configurable, is "admin" here for providing a default auth source or because one is required in the url but it will then be overridden by the properties or both?
I don't see it in the "connection-fields" below though, is that on purpose?

connector/connectionFields.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-fields>
+  <field name="server" label="Server" value-type="string" category="endpoint" />
+
+  <field name="port" label="Port" value-type="string" category="endpoint" default-value="27017" />
+
+  <field name="authentication" label="Authentication" category="authentication" value-type="string" editable="false" default-value="auth-user-pass" />
```
nbagnard, line 8: I guess this is here to lay the foundation for allowing different auth types?
What's the vision here? Will you convert this to a drop box and then use conditional display for the fields to provide based on the auth type later?

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- See notes below for definitions that require 2020.3 or newer -->
+<dialect name='FullPostgreSQL90JDBC'
```
nbagnard, line 4: Maybe we could call it "MongoSQL" instead of "FullPostgreSQL90JDBC"?

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+         class='mongosql_jdbc'
+         version='18.1'>
```
nbagnard, line 6: What's the version here? Is it tied to something? Is it necessary?

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+  <function-map>
```
nbagnard, line 7: Where do the function come from? Do we have to provide a "translation" for a set of functions Tableau needs and these are the functions or is this a list of function that the language supports? 
Did you check these function against:
- the function list in the driver
-  MongoSQL
?
It does not seem to quite match the function list in the driver at least and some "translation" are looking odd.

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+    <function group='numeric' name='ABS' return-type='real'>
```
nbagnard, line 8: From what I can tell (MongoSQL syntax and driver code)ABS is not supported in MongoSQL.

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+      <formula>ABS(%1)</formula>
+      <argument type='real' />
+    </function>
+    <function group='numeric' name='ABS' return-type='int'>
+      <formula>ABS(%1)</formula>
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='ACOS' return-type='real'>
```
nbagnard, line 16: I also don't see ACOS, ASIN, ATAN, ATAN2, CEILING, COS, COT, DEGREES, DIV, EXP, FLOOR, HEXBINX and HEXBINY use ABS, etc...
If we have a translation, then, it's Ok to include them, but it looks like a lot of them here are relying on the function being present in the DB language.

This is the list of functions and their argument for MongoSQL from the driver (null means any type):
BIT_LENGTH(string)
CAST(null, null, null, null)
CHAR_LENGTH(string)
COALESCE( null)
CURRENT_TIMESTAMP()
CURRENT_TIMESTAMP(int)
EXTRACT(string)
LOWER(string)
NULLIF(null, null)
OCTET_LENGTH(string)
POSITION(string, string)
SLICE(array, int,int)
SIZE(array)
SUBSTRING(string, long)
SUBSTRING(string, long, long)
TRIM(string)
TRIM(string,string)
TRIM(string,string,string)
UPPER(string)

Also, note that the syntax is not always just the function name and its arguments, for example EXTRACT requires the field name. EXTRACT "(" <extract field> FROM [<expression>] ")" 
<extract field> ::= TIMEZONE_HOUR | TIMEZONE_MINUTE
                  | YEAR | MONTH | DAY | HOUR | MINUTE | SECOND
For the complete spec, you can look into the "Grammar" section of "Scalar Functions" in the MongoSQL Syntax doc, I think it is needed to have the most accurate list.

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+      <argument type='localstr' />
+      <argument type='datetime' />
+      <argument type='localstr' />
+    </date-function>
+    <native-split-function>
+      <formula part='left'>SPLIT_PART(%1, %2, %3)</formula>
+      <formula part='right'>REVERSE(SPLIT_PART(REVERSE(%1),REVERSE(%2),%3*-1))</formula>
+    </native-split-function>
+  </function-map>
+  <supported-aggregations>
```
nbagnard, line 1027: You probably also want to double-check this list.

connector/dialect.tdd
```diff
@@ -0,0 +1,1258 @@
...
+    <aggregation value='TRUNC_HOUR'/>
+    <aggregation value='TRUNC_MINUTE'/>
+    <aggregation value='TRUNC_SECOND'/>
+    <aggregation value='TRUNC_ISO_YEAR'/>
+    <aggregation value='TRUNC_ISO_QTR'/>
+    <aggregation value='TRUNC_ISO_WEEK'/>
+    <aggregation value='TRUNC_ISO_WEEKDAY'/>
+  </supported-aggregations>
+  <sql-format>
+    <date-literal-escape value='PostgresStyle' />
```
nbagnard, line 1074: What does PostresStyle means? Do you know?

connector/connectionBuilder.js
bucaojit: The connector files that are in this PR are mainly from the sample plugin with a few minor changes to allow the connection to ADL work and have a package output `mongosql_jdbc.taco` for the evergreen build.  

Further customization is in SQL-733 where things like the authSource and other fields are added.

connector/connectionFields.xml
bucaojit: Yes, the different authentication methods can be set here: user/pass, OAuth, Integrated (for SSO).  I see that ADL supports [user/pass ](https://docs.mongodb.com/datalake/tutorial/connect/) and was going to stick with that for now.

nbagnard (approved):
LGTM! 👍 
Sorry for the slew of comments, I misunderstood the scope of the PR

connector/connectionProperties.js
bucaojit: Removed `compressors` as it is ignored by ADL.  `authSource` is used, keeping that property.

bucaojit:
evergreen merge

PR #3, opened 2022-02-23
SQL-733: Implemented custom Tableau Connector files

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-733

Implementing the files other than the Capabilities section of the `manifest.xml` and Dialect definition file `dialect.tdd`.

### DIFF
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -21,27 +21,6 @@
 
   <field name="v-authDatabase" label="Authentication Database" value-type="string" category="authentication" default-value="admin" />
 
-  <field name="v-loglevel" label="JDBC Log Level" value-type="selection" default-value="log-info" category="general" >
-    <selection-group>
-      <option value="log-debug" label="DEBUG" />
-      <option value="log-info" label="INFO" />
-      <option value="log-error" label="ERROR" />
-    </selection-group>
-  </field>
-
-  <field name="v-logdiroption" label="Custom JDBC Log Directory" value-type="boolean" category="general"  >
-    <boolean-options>
-      <false-value value="" />
-      <true-value value="required" />
-    </boolean-options>
-  </field>
-
-  <field name="v-log-directory" label="JDBC Log Directory" value-type="string" default-value="" category="general"  >
-    <conditions>
-      <condition field="v-logdiroption" value="required"/>
-    </conditions>
-  </field>
-
   <field name="sslmode" label="Require SSL" value-type="boolean" category="general" default-value="" >
     <boolean-options>
       <false-value value="" />
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -10,11 +10,4 @@
 
   <field name="password" label="Password" value-type="string" category="authentication" secure="true" />
 
-  <field name="sslmode" label="Require SSL" value-type="boolean" category="general" default-value="" >
-    <boolean-options>
-      <false-value value="" />
-      <true-value value="require" />
-    </boolean-options>
-  </field>
-
 </connection-fields>
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -2,7 +2,7 @@
 
 <connection-fields>
 
-  <field name="v-connection-uri" label="Connection URI" value-type="string" category="endpoint" />
+  <field name="v-mongodb-uri" label="MongoDB URI" value-type="string" category="endpoint" />
 
   <field name="authentication" label="Authentication" category="authentication" value-type="string" editable="false" default-value="auth-user-pass" />
 
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -2,24 +2,13 @@
 
 <connection-fields>
 
-  <field name="v-protocol" label="Connection Protocol" value-type="selection" category="endpoint" >
-    <selection-group>
-      <option value="mongodb" label="mongodb" />
-      <option value="mongodb+srv" label="mongodb+srv" />
-    </selection-group>
-  </field>
-
-  <field name="server" label="Server" value-type="string" category="endpoint" />
-
-  <field name="port" label="Port" value-type="string" category="endpoint" default-value="27017" />
+  <field name="v-connection-uri" label="Connection URI" value-type="string" category="endpoint" />
 
   <field name="authentication" label="Authentication" category="authentication" value-type="string" editable="false" default-value="auth-user-pass" />
 
   <field name="username" label="Username" value-type="string" category="authentication" />
 
   <field name="password" label="Password" value-type="string" category="authentication" secure="true" />
-
-  <field name="v-authDatabase" label="Authentication Database" value-type="string" category="authentication" default-value="admin" />
 
   <field name="sslmode" label="Require SSL" value-type="boolean" category="general" default-value="" >
     <boolean-options>
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-fields>
+  <field name="v-prefix" label="Connection Prefix" value-type="selection" category="endpoint" >
+    <selection-group>
+      <option value="mongodb" label="mongodb" />
+      <option value="mongodb+srv" label="mongodb+srv" />
+    </selection-group>
+  </field>
+
+  <field name="server" label="Server" value-type="string" category="endpoint" />
+
+  <field name="port" label="Port" value-type="string" category="endpoint" default-value="27017" />
+
+  <field name="authentication" label="Authentication" category="authentication" value-type="string" editable="false" default-value="auth-user-pass" />
+
+  <field name="username" label="Username" value-type="string" category="authentication" />
+
+  <field name="password" label="Password" value-type="string" category="authentication" secure="true" />
+
+  <field name="v-authDatabase" label="Authentication Database" value-type="string" category="authentication" default-value="admin" />
+
+  <field name="v-loglevel" label="JDBC Log Level" value-type="selection" default-value="log-info" category="general" >
+    <selection-group>
+      <option value="log-debug" label="DEBUG" />
+      <option value="log-info" label="INFO" />
+      <option value="log-error" label="ERROR" />
+    </selection-group>
+  </field>
+
+  <field name="v-logdiroption" label="Custom JDBC Log Directory" value-type="boolean" category="general"  >
+    <boolean-options>
+      <false-value value="" />
+      <true-value value="required" />
+    </boolean-options>
+  </field>
+
+  <field name="v-log-directory" label="JDBC Log Directory" value-type="string" default-value="" category="general"  >
+    <conditions>
+      <condition field="v-logdiroption" value="required"/>
+    </conditions>
+  </field>
+
+  <field name="sslmode" label="Require SSL" value-type="boolean" category="general" default-value="" >
+    <boolean-options>
+      <false-value value="" />
+      <true-value value="require" />
+    </boolean-options>
+  </field>
+
+</connection-fields>
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -2,7 +2,7 @@
 
 <connection-fields>
 
-  <field name="v-prefix" label="Connection Prefix" value-type="selection" category="endpoint" >
+  <field name="v-protocol" label="Connection Protocol" value-type="selection" category="endpoint" >
     <selection-group>
       <option value="mongodb" label="mongodb" />
       <option value="mongodb+srv" label="mongodb+srv" />
--- a/connector/connectionMetadata.xml
+++ b/connector/connectionMetadata.xml
@@ -2,7 +2,7 @@
 
 <connection-metadata>
   <database enabled='true' label='Database'>
-    <field optional='true' />
+    <field optional='false' />
   </database>
   <schema enabled='false' />
   <table enabled='true' label='Table' />
--- a/connector/connectionMetadata.xml
+++ b/connector/connectionMetadata.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-metadata>
+  <database enabled='true' label='Database'>
+    <field optional='true' />
+  </database>
+  <schema enabled='false' />
+  <table enabled='true' label='Table' />
+</connection-metadata>
--- a/connector/connectionProperties.js
+++ b/connector/connectionProperties.js
@@ -1,0 +1,26 @@
+(function propertiesbuilder(attr) {
+    var props = {};
+    props["user"] = attr[connectionHelper.attributeUsername];
+    props["password"] = attr[connectionHelper.attributePassword];
+    props["authSource"] = attr["v-authDatabase"];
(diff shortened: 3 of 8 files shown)


connector/connectionFields.xml
```diff
@@ -1,6 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <connection-fields>
+
```
bucaojit, line 4: Screenshot of the connection dialogue [here](https://docs.google.com/document/d/1vt6f0c2T74jgb_QqLFM8FEM12dh56BfD2Qh5Gw6rxWc/edit#bookmark=id.oc2z3uwg2ulz)

connector/connectionResolver.tdr
```diff
@@ -1,24 +1,11 @@
 <?xml version='1.0' encoding='utf-8' ?>
 <tdr class='mongosql_jdbc'>
-	<connection-resolver>
-    <connection-builder>
-      <script file='connectionBuilder.js'/>
-    </connection-builder>
-    <connection-normalizer>
```
bucaojit, line 7: Removing connection-normalizer to utilize inferred connection resolver.

connector/connectionFields.xml
```diff
@@ -1,6 +1,14 @@
...
+  <field name="v-prefix" label="Connection Prefix" value-type="selection" category="endpoint" >
```
rychipman: It seems like "protocol" might be a more appropriate name, since that's what `mongodb` and `mongodb+srv` are

connector/connectionFields.xml
```diff
@@ -11,6 +19,29 @@
...
+      <false-value value="" />
+      <true-value value="required" />
+    </boolean-options>
+  </field>
+
+  <field name="v-log-directory" label="JDBC Log Directory" value-type="string" default-value="" category="general"  >
+    <conditions>
+      <condition field="v-logdiroption" value="required"/>
+    </conditions>
+  </field>
```
rychipman:
```suggestion
```
suggestion: can we leave this out for now and talk further in the design doc about whether we want it? This seems like it might be a bit too niche to merit inclusion in the UI itself, but I think we should defer it for further discussion

connector/connectionFields.xml
```diff
@@ -11,6 +19,29 @@
 
   <field name="password" label="Password" value-type="string" category="authentication" secure="true" />
 
+  <field name="v-authDatabase" label="Authentication Database" value-type="string" category="authentication" default-value="admin" />
```
rychipman: I think we should talk about the general stragegy here wrt various MongoDB URI parameters. Specifically, the way this patch is going seems to suggest that any URI parameter we want to enable will require a separate configuration UI element to be added.

I'm curious to know whether/why this is preferable to just allowing the user to provide the whole MongoDB URI

connector/manifest.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version='1.0' encoding='utf-8' ?>
 
-<connector-plugin class='mongosql_jdbc' superclass='jdbc' plugin-version='0.0.0' name='MongoSQL' version='18.1' min-version-tableau='2020.4'>
+<connector-plugin class='mongosql_jdbc' superclass='jdbc' plugin-version='0.1.0' name='MongoSQL' version='18.1' min-version-tableau='2021.1'>
```
rychipman: Let's talk with Alexi about this, but I don't _think_ we want to use "MongoSQL" here. Instead, I think "Atlas SQL" or "MongoDB" will be preferred, depending on the context

connector/manifest.xml
```diff
@@ -1,14 +1,14 @@
...
   <vendor-information>
       <company name="MongoDB"/>
-      <support-link url="https://example.com"/>
-      <driver-download-link url="https://drivers.example.com"/>
+      <support-link url="https://support.mongodb.com"/>
+      <driver-download-link url="https://search.maven.org/search?q=a:mongodb-jdbc"/>
```
rychipman:
```suggestion
      <driver-download-link url="https://search.maven.org/artifact/org.mongodb/mongodb-jdbc"/>
```

connector/connectionProperties.js
```diff
@@ -2,9 +2,20 @@
     var props = {};
     props["user"] = attr[connectionHelper.attributeUsername];
     props["password"] = attr[connectionHelper.attributePassword];
+    props["authSource"] = attr["v-authDatabase"];
+    props["database"] = attr[connectionHelper.attributeDatabase];
     props["dialect"] = "mongosql";
-    props["database"] = attr[connectionHelper.attributeDatabase]
-    props["authSource"] = "admin";
+    props["TYPE_MAPPING_MODE"] = "tableau";
```
rychipman:
```suggestion
```
I don't think this property exists in the JDBD driver, does it?

connector/manifest.xml
rychipman: Another thought/question: where does this download link appear in the connector? Is it intended for users to follow and download manually, or is it used automatically by Tableau somehow?

connector/manifest.xml
```diff
@@ -1,14 +1,14 @@
...
   </vendor-information>
   <connection-customization class="mongosql_jdbc" enabled="true" version="10.0">
-    <vendor name="vendor"/>
-    <driver name="driver"/>
+    <vendor name="MongoDB"/>
+    <driver name="MongoSQL Connector"/>
```
rychipman: do you know where these values actually get used by the connector? I couldn't find any explanation in the docs

connector/connectionMetadata.xml
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <connection-metadata>
-  <database enabled='true'>
-    <field />
+  <database enabled='true' label='Database'>
+    <field optional='true' />
```
rychipman: What happens if the user doesn't provide a value for the database? Don't many of the JDBC driver's API calls require a database to be set in order to work?

rychipman (changes requested):
Great start! some small suggestions and a bunch of questions from me

connector/connectionFields.xml
```diff
@@ -11,6 +19,29 @@
...
+
+  <field name="v-loglevel" label="JDBC Log Level" value-type="selection" default-value="log-info" category="general" >
```
nbagnard: Shall we wait for logging to be implemented and available before adding this to the dialog? It seems a bit odd to have it in the dialog before it actually is available.

connector/manifest.xml
nbagnard: Are we planning to also provide a connector bundle for ODBC at some point?
If so, having a JDBC suffix or prefix would make it easier for us in the future to add the ODBC connector and avoid confusion.
Also as a side note, I personally appreciate when a connector tells me specifically which kind of driver it is using underneath. It might not be as useful for non technical users, but the ones familiar with ODBC and JDBC will likely be happy to know as well.

connector/connectionFields.xml
nbagnard: I share some of the reserve Ryan has here. It would be nice to be able to let the user also input his custom jdbc url, because otherwise we are restricting them to only the parameters that are in the dialog. These here are the ones we use, but other users can have more refined urls best suited for their needs. 
The issue I see is that not all the parameters are supported in the jdbc url. dialect isn't (but it's not visible for Tableau users, so not a blocker), logging params won't be though unless we add support for them in the jdbc driver itself. 
In short, if not all parameters are supported in the url, we can't rely on only letting the user provide its url. Maybe a conditional display which only "asks" for the parameters which are not supported in the url when opting for the non default url would do the trick? 
I'm also (still) very much in-favour of supporting all parameters as part of the jdbc url ;)

connector/manifest.xml
nbagnard: Is _min-version-tableau='2021.1'_ not going to be too restrictive and cut us from potential users who are using a slightly old version of Tableau?
Was _2020.4_ looking to old?

connector/manifest.xml
bucaojit: I see this download link show up in the error dialogue when the jdbc driver isn't present in the Tableau/Drivers/ directory.  It's meant for the user to follow and install manually.  With the following text:

No suitable driver installed, or the URL is incorrect
Connector Class: mongosql_jdbc, Version: 0.1.0
[Download and install the drivers, and then connect.](https://search.maven.org/artifact/org.mongodb/mongodb-jdbc)
[For support, contact MongoDB.](https://support.mongodb.com/)

connector/connectionProperties.js
bucaojit: It is in the design for the type mapping, I will remove it for now.

connector/manifest.xml
bucaojit: I don't see where it is used either, it doesn't show up in the logs.  The fields are needed to pass the packager validation however.

connector/connectionMetadata.xml
bucaojit: That's a good point.  If the field isn't set, I see Tableau is able to connect but the user needs to choose the database from the drop-down menu.  In the logs I see Tableau making a  `MongoDatabaseMetaData` call to get the database names.  
I can change the database field to being required.

connector/manifest.xml
nbagnard: Weird, their doc says "Vendor defined attributes will be logged and persisted to Tableau workbook xml in plain text." https://tableau.github.io/connector-plugin-sdk/docs/connection-dialog
It also seems that they can be used as connection attributes, but I don't think we need them, do we?

bucaojit: The vendor defined attributes referred to there are the ones defined in the connectionFields.xml that start with the `v-` like the `v-loglevel` for example.
The fields in connectionFields.xml are the ones that would be used for the `<required-attributes>` element, which we wouldn't need if we use the inferred connection resolver, added in the 2021.1 release.

connector/connectionFields.xml
bucaojit: Sure, it's removed.

connector/manifest.xml
bucaojit: For Tableau there aren't plans to provide a connector for ODBC.  
In the "Installed Connectors" section the connector names are generally the database name.  I do see some connectors in the "Additional Connectors" that show JDBC or ODBC, I'm not sure yet why there are the two sections.

bucaojit: Good question, the 2021.1 version allows for the inferred connection resolver, which is used in the connector however may not be necessary.  I'll test out with setting the resolver and checking again if there is a benefit.

nbagnard: Sounds good.
I guess if there is only flavour of the connector then it does make sense to not have JDBC/ODBC.
When I say "at some point", I looking at a few years from now. If we know that we will never want to provide a connector using the ODBC driver, that's fine. And if we do, then we'll find a differentiator at this time :)

connector/manifest.xml
nbagnard: I see.
Well, then it will stay a mystery. It could also be something related to a Tableau server user case. Who knows...

connector/connectionFields.xml
bucaojit: Spoke with Ryan offline and as the current iteration has few properties to set such as 'database', I will make the change to allow the user to provide the mongodb URI and for the additional properties we can discuss the design of when we get to it.

connector/connectionResolver.tdr
bucaojit: Re-added to work with older version 2020.4

connector/connectionFields.xml
bucaojit: Removed this field as we are using connection URI

connector/manifest.xml
bucaojit: I added back the `connection-normalizer` and tested with 2020.4 and it worked fine.  2020.3 is the first version that uses the Connection Dialogs V2, but that version is no longer available due to the log4j vulnerability so `2020.4` is a good min version.

bucaojit:
Updated the connection [dialogue screenshot](https://docs.google.com/document/d/1vt6f0c2T74jgb_QqLFM8FEM12dh56BfD2Qh5Gw6rxWc/edit#bookmark=id.oc2z3uwg2ulz)

Checked with Alexi on the naming, updated name to "MongoDB Atlas".

mattChiaravalloti (approved):

connector/connectionFields.xml
```diff
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <connection-fields>
-  <field name="server" label="Server" value-type="string" category="endpoint" />
 
-  <field name="port" label="Port" value-type="string" category="endpoint" default-value="27017" />
+  <field name="v-connection-uri" label="Connection URI" value-type="string" category="endpoint" />
```
rychipman:
```suggestion
  <field name="v-mongodb-uri" label="MongoDB URI" value-type="string" category="endpoint" />
```
nit: let's change this name to make it very clear that this is a MongoDB URI, not some other URI

connector/manifest.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version='1.0' encoding='utf-8' ?>
 
-<connector-plugin class='mongosql_jdbc' superclass='jdbc' plugin-version='0.0.0' name='MongoSQL' version='18.1' min-version-tableau='2020.4'>
+<connector-plugin class='mongosql_jdbc' superclass='jdbc' plugin-version='0.1.0' name='MongoDB Atlas' version='18.1' min-version-tableau='2020.4'>
```
rychipman:
```suggestion
<connector-plugin class='mongodb_jdbc' superclass='jdbc' plugin-version='0.1.0' name='MongoDB Atlas' version='18.1' min-version-tableau='2020.4'>
```
let's change this "mongosql" to "mongodb" as well

connector/connectionProperties.js
```diff
@@ -2,9 +2,8 @@
     var props = {};
     props["user"] = attr[connectionHelper.attributeUsername];
     props["password"] = attr[connectionHelper.attributePassword];
+    props["database"] = attr[connectionHelper.attributeDatabase];
     props["dialect"] = "mongosql";
-    props["database"] = attr[connectionHelper.attributeDatabase]
-    props["authSource"] = "admin";
 
     if (attr[connectionHelper.attributeSSLMode] == "require") {
         props["ssl"] = "true";
```
rychipman: Are SSL properties currently handled at all by the JDBC driver? I don't see any tests asserting that we handle them in `MongoDriverTest.java`. Beyond that, I think we also have additional questions to answer around how to handle conflicts between SSL options in the properties and in the URI.

If there's not a clear and obvious answer to these things at the moment, I would suggest that we remove this for now and create a new design doc section to discuss, so that this question doesn't hold up merging the rest of the PR

rychipman (changes requested):
Just a few small things remaining, otherwise looks great!

connector/connectionProperties.js
bucaojit: I see that the `ssl` property gets added to the ConnectionString in the driver when I print the connection string.  I have also noticed that whether that property is set or not affects whether the driver can connect to ADL, with the local ADL instance it only works if unset.

I will remove the SSL UI element and users can specify it in the MongoDB URI

rychipman (approved):
LGTM!

nbagnard (approved):
LGTM!

bucaojit:
evergreen merge

PR #4, opened 2022-03-02
SQL-734: Implemented capabilities section of manifest.xml

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-734

Details of options in [design doc ](https://docs.google.com/document/d/1vt6f0c2T74jgb_QqLFM8FEM12dh56BfD2Qh5Gw6rxWc/edit#bookmark=id.kldg82b80epj)

Full list of capabilities [here](https://tableau.github.io/connector-plugin-sdk/docs/capabilities)

### DIFF
--- a/connector/manifest.xml
+++ b/connector/manifest.xml
@@ -25,16 +25,18 @@
       <!-- Query -->
       <customization name="CAP_QUERY_BOOL_IDENTIFIER_TO_LOGICAL" value="yes"/>
       <customization name="CAP_QUERY_BOOLEXPR_TO_INTEXPR" value="no"/>
-      <customization name="CAP_QUERY_CASE_OUT_NO_BOOL_OPS" value="yes"/>
-      <customization name="CAP_QUERY_GROUP_BY_ALIAS" value="yes"/>
+      <customization name="CAP_QUERY_CASE_OUT_NO_BOOL_OPS" value="no"/>
+      <customization name="CAP_QUERY_GROUP_BY_ALIAS" value="no"/>
       <customization name="CAP_QUERY_GROUP_BY_BOOL" value="yes"/>
       <customization name="CAP_QUERY_HAVING_UNSUPPORTED" value="no"/>
-      <customization name="CAP_QUERY_INCLUDE_GROUP_BY_COLUMNS_IN_SELECT" value="yes"/>
-      <customization name="CAP_QUERY_JOIN_PUSH_DOWN_CONDITION_EXPRESSIONS" value="yes"/>
+      <customization name="CAP_QUERY_INCLUDE_GROUP_BY_COLUMNS_IN_SELECT" value="no"/>
+      <customization name="CAP_QUERY_JOIN_PUSH_DOWN_CONDITION_EXPRESSIONS" value="no"/>
+      <customization name="CAP_QUERY_SCALAR_SELECTS_ALL_IN_GROUP_BYS" value="yes"/>
       <customization name="CAP_QUERY_SORT_BY" value="yes"/>
       <customization name="CAP_QUERY_SUBQUERIES" value="yes"/>
       <customization name="CAP_QUERY_SUPPORT_EMPTY_GROUPBY" value="no"/>
       <customization name="CAP_QUERY_TOP_0" value="no"/>
+      <customization name="CAP_QUERY_TOP_N" value="yes"/>
       <!-- JDBC -->
       <customization name="CAP_JDBC_QUERY_CANCEL" value="no"/>
       <customization name="CAP_JDBC_SUPPRESS_ENUMERATE_SCHEMAS" value="yes"/>
--- a/connector/manifest.xml
+++ b/connector/manifest.xml
@@ -10,8 +10,8 @@
     <vendor name="vendor"/>
     <driver name="driver"/>
     <customizations>
+      <!-- https://tableau.github.io/connector-plugin-sdk/docs/capabilities -->
       <!-- Metadata -->
-      <customization name="CAP_FAST_METADATA" value="yes"/>
       <customization name="CAP_QUERY_WHERE_FALSE_METADATA" value="yes"/>
       <!-- Temporary Tables -->
       <customization name="CAP_CREATE_TEMP_TABLES" value="no"/>
@@ -23,21 +23,16 @@
       <!-- Initial SQL-->
       <customization name="CAP_SUPPORTS_INITIAL_SQL" value="yes"/>
       <!-- Query -->
-      <customization name="CAP_QUERY_ALLOW_JOIN_REORDER" value="yes"/>
       <customization name="CAP_QUERY_BOOL_IDENTIFIER_TO_LOGICAL" value="yes"/>
       <customization name="CAP_QUERY_BOOLEXPR_TO_INTEXPR" value="no"/>
       <customization name="CAP_QUERY_CASE_OUT_NO_BOOL_OPS" value="yes"/>
       <customization name="CAP_QUERY_GROUP_BY_ALIAS" value="yes"/>
       <customization name="CAP_QUERY_GROUP_BY_BOOL" value="yes"/>
-      <customization name="CAP_QUERY_GROUP_BY_DEGREE" value="no"/>
       <customization name="CAP_QUERY_HAVING_UNSUPPORTED" value="no"/>
       <customization name="CAP_QUERY_INCLUDE_GROUP_BY_COLUMNS_IN_SELECT" value="yes"/>
       <customization name="CAP_QUERY_JOIN_PUSH_DOWN_CONDITION_EXPRESSIONS" value="yes"/>
       <customization name="CAP_QUERY_SORT_BY" value="yes"/>
-      <customization name="CAP_QUERY_SORT_BY_DEGREE" value="yes"/>
       <customization name="CAP_QUERY_SUBQUERIES" value="yes"/>
-      <customization name="CAP_QUERY_SUBQUERIES_WITH_TOP" value="yes"/>
-      <customization name="CAP_QUERY_SUBQUERY_QUERY_CONTEXT" value="yes"/>
       <customization name="CAP_QUERY_SUPPORT_EMPTY_GROUPBY" value="no"/>
       <customization name="CAP_QUERY_TOP_0" value="no"/>
       <!-- JDBC -->
--- a/connector/manifest.xml
+++ b/connector/manifest.xml
@@ -10,24 +10,39 @@
     <vendor name="vendor"/>
     <driver name="driver"/>
     <customizations>
-      <customization name="CAP_SELECT_INTO" value="yes"/>
-      <customization name="CAP_SELECT_TOP_INTO" value="yes"/>
-      <customization name="CAP_CREATE_TEMP_TABLES" value="no"/>
-      <customization name="CAP_QUERY_BOOLEXPR_TO_INTEXPR" value="no"/>
-      <customization name="CAP_QUERY_GROUP_BY_BOOL" value="yes"/>
-      <customization name="CAP_QUERY_GROUP_BY_DEGREE" value="yes"/>
-      <customization name="CAP_QUERY_SORT_BY" value="yes"/>
-      <customization name="CAP_QUERY_SUBQUERIES" value="yes"/>
-      <customization name="CAP_QUERY_TOP_N" value="yes"/>
-      <customization name="CAP_QUERY_TOP_SAMPLE" value="yes"/>
-      <customization name="CAP_QUERY_TOP_SAMPLE_PERCENT" value="yes"/>
+      <!-- Metadata -->
+      <customization name="CAP_FAST_METADATA" value="yes"/>
       <customization name="CAP_QUERY_WHERE_FALSE_METADATA" value="yes"/>
+      <!-- Temporary Tables -->
+      <customization name="CAP_CREATE_TEMP_TABLES" value="no"/>
+      <customization name="CAP_SELECT_INTO" value="no"/>
+      <customization name="CAP_SELECT_TOP_INTO" value="no"/>
+      <!-- String Splits -->
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_LEFT" value="no"/>
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_RIGHT" value="no"/>
+      <!-- Initial SQL-->
+      <customization name="CAP_SUPPORTS_INITIAL_SQL" value="yes"/>
+      <!-- Query -->
+      <customization name="CAP_QUERY_ALLOW_JOIN_REORDER" value="yes"/>
+      <customization name="CAP_QUERY_BOOL_IDENTIFIER_TO_LOGICAL" value="yes"/>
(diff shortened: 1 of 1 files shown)


connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
     <vendor name="vendor"/>
     <driver name="driver"/>
     <customizations>
-      <customization name="CAP_SELECT_INTO" value="yes"/>
-      <customization name="CAP_SELECT_TOP_INTO" value="yes"/>
+      <!-- Metadata -->
```
rychipman, line 14: suggestion: could we add a comment to the top of this capability list that includes a link to the [capabilities documentation](https://tableau.github.io/connector-plugin-sdk/docs/capabilities)?

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <customization name="CAP_FAST_METADATA" value="yes"/>
```
rychipman: question: What's our approach here when we want to keep the default value? Do we explicitly set to the default, or are we just omitting it?

I see that we are explicitly setting the default for `CAP_FAST_METADATA`, but implicitly using the default for `CAP_QUERY_TOP_0_METADATA` by omission. Unless there's a reason to do otherwise, I'd suggest that we take a consistent approach across all capabilities

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <customization name="CAP_QUERY_WHERE_FALSE_METADATA" value="yes"/>
+      <!-- Temporary Tables -->
       <customization name="CAP_CREATE_TEMP_TABLES" value="no"/>
+      <customization name="CAP_SELECT_INTO" value="no"/>
+      <customization name="CAP_SELECT_TOP_INTO" value="no"/>
+      <!-- String Splits -->
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_LEFT" value="no"/>
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_RIGHT" value="no"/>
```
rychipman, line 22: perhaps we could file a ticket to support string splitting, since we know tableau will use it if it exists?

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <!-- Initial SQL-->
+      <customization name="CAP_SUPPORTS_INITIAL_SQL" value="yes"/>
+      <!-- Query -->
+      <customization name="CAP_QUERY_ALLOW_JOIN_REORDER" value="yes"/>
+      <customization name="CAP_QUERY_BOOL_IDENTIFIER_TO_LOGICAL" value="yes"/>
```
rychipman, line 26: the documentation for this capability is phrased _very_ confusingly. The sentence as written in the docs suggests that "yes" means "don't wrap boolean predicates in an equality expression", but the naming of the capability seems to suggest the opposite

perhaps we could investigate to confirm one way or the other? I don't think it's a big enough deal to hold up this whole PR, but filing a ticket to investigate it might be warranted

mattChiaravalloti (changes requested):
Didn't inspect these too closely, but I'm assuming they match the table in the design doc. I left several comments on there that we should discuss before I LGTM!

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
       <customization name="CAP_QUERY_BOOLEXPR_TO_INTEXPR" value="no"/>
+      <customization name="CAP_QUERY_CASE_OUT_NO_BOOL_OPS" value="yes"/>
```
rychipman: what can't our case outputs contain boolean ops? I would have thought we'd support any arbitrary expression in a case branch

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <customization name="CAP_QUERY_GROUP_BY_ALIAS" value="yes"/>
```
rychipman: I don't _think_ we support this, do we?

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
       <customization name="CAP_QUERY_GROUP_BY_BOOL" value="yes"/>
-      <customization name="CAP_QUERY_GROUP_BY_DEGREE" value="yes"/>
+      <customization name="CAP_QUERY_GROUP_BY_DEGREE" value="no"/>
+      <customization name="CAP_QUERY_HAVING_UNSUPPORTED" value="no"/>
+      <customization name="CAP_QUERY_INCLUDE_GROUP_BY_COLUMNS_IN_SELECT" value="yes"/>
```
rychipman: Is this required? I think it should be perfectly fine to have a group by column that's not referenced by the SELECT

rychipman (changes requested):
haven't quite gotten through all of these yet, but have to step away. Figured I'd post what I have so far instead of waiting to come back later, since all the capabilities are orthogonal

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <customization name="CAP_QUERY_JOIN_PUSH_DOWN_CONDITION_EXPRESSIONS" value="yes"/>
```
rychipman: This is another confusingly named attribute, but the documentation description suggests that "yes" here means "doesn't support arbitrary exprs in joins", which I don't think is true!

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
       <customization name="CAP_QUERY_SORT_BY" value="yes"/>
```
rychipman, line 35: should we have `CAP_QUERY_SCALAR_SELECTS_ALL_IN_GROUP_BYS` set to "yes" here?

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
-      <customization name="CAP_QUERY_WHERE_FALSE_METADATA" value="yes"/>
       <customization name="CAP_QUERY_SUBQUERIES_WITH_TOP" value="yes"/>
-      <customization name="CAP_SUPPORTS_SPLIT_FROM_LEFT" value="yes"/>
-      <customization name="CAP_SUPPORTS_SPLIT_FROM_RIGHT" value="yes"/>
-      <customization name="CAP_SUPPORTS_UNION" value="yes"/>
-      <customization name="CAP_QUERY_ALLOW_PARTIAL_AGGREGATION" value="no"/>
-      <customization name="CAP_QUERY_TIME_REQUIRES_CAST" value="yes"/>
+      <customization name="CAP_QUERY_SUBQUERY_QUERY_CONTEXT" value="yes"/>
+      <customization name="CAP_QUERY_SUPPORT_EMPTY_GROUPBY" value="no"/>
+      <customization name="CAP_QUERY_TOP_0" value="no"/>
```
rychipman, line 38: I think mongosql _should_ support this. Also, should we be setting `CAP_QUERY_TOP_N` to "yes"?

connector/manifest.xml
```diff
@@ -10,24 +10,39 @@
...
+      <!-- JDBC -->
+      <customization name="CAP_JDBC_QUERY_CANCEL" value="no"/>
```
rychipman, line 43: Perhaps we should file a ticket for this one? I think we _could_ support it in theory, even if we don't now

rychipman (changes requested):
put up comments on the rest of the capabilities!

connector/manifest.xml
bucaojit: Sure, created [SQL-750](https://jira.mongodb.org/browse/SQL-750) to support string splitting.

connector/manifest.xml
bucaojit: Made the change to omit the option where I was setting the values that were the same as the default.  I tested it and confirmed the behavior was the same.  
I was following the document too literally where it said "set 'yes' if..." even where 'yes' was the default.  Things are consistent now.

connector/manifest.xml
bucaojit: I'll switch this to `no`.  I did not find a CASE query with `AND/OR/NOT` in the tests and had thought it was not supported.

connector/manifest.xml
bucaojit: Yes, I missed the `CAP_QUERY_TOP_N`, I'll add that!

I took a look at why the `limit 0` was failing, looks like a [bug in mongoast](https://github.com/10gen/mongoast/blob/efb8f399b32f01e904698f89589a64adece4af87/parser/parse_stage.go#L153) that fails if the argument is 0.

rychipman: Ah, looks like that might be intentional, and I'm the one who's confused here. While MongoSQL doesn't do anything to disallow `LIMIT 0`, it looks like `$limit` requires a positive number in MongoDB agg.

connector/manifest.xml
bucaojit: Created investigation ticket [SQL-751](https://jira.mongodb.org/browse/SQL-751)

connector/manifest.xml
bucaojit: Filed ticket to address this [SQL-752](https://jira.mongodb.org/browse/SQL-752)

connector/manifest.xml
bucaojit: I looked into this further and the case I was looking at had a table alias, not a column alias.  Switching this to `no`.

connector/manifest.xml
```diff
@@ -10,24 +10,36 @@
...
+      <customization name="CAP_SELECT_TOP_INTO" value="no"/>
+      <!-- String Splits -->
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_LEFT" value="no"/>
+      <customization name="CAP_SUPPORTS_SPLIT_FROM_RIGHT" value="no"/>
+      <!-- Initial SQL-->
+      <customization name="CAP_SUPPORTS_INITIAL_SQL" value="yes"/>
+      <!-- Query -->
+      <customization name="CAP_QUERY_BOOL_IDENTIFIER_TO_LOGICAL" value="yes"/>
       <customization name="CAP_QUERY_BOOLEXPR_TO_INTEXPR" value="no"/>
+      <customization name="CAP_QUERY_CASE_OUT_NO_BOOL_OPS" value="no"/>
```
mattChiaravalloti, line 29: In the same spirit of this capability, the `CAP_QUERY_CASE_MATCHES_NULL` description seems to describe behavior that is supported by MongoSQL, but it defaults to "no". Did you investigate that capability as part of this ticket already and determine the "no" is the correct value for us?

connector/manifest.xml
```diff
@@ -10,24 +10,36 @@
...
+      <customization name="CAP_QUERY_GROUP_BY_ALIAS" value="no"/>
```
mattChiaravalloti: The default for this one is "no" so should we omit it?

connector/manifest.xml
```diff
@@ -10,24 +10,36 @@
...
       <customization name="CAP_QUERY_GROUP_BY_BOOL" value="yes"/>
-      <customization name="CAP_QUERY_GROUP_BY_DEGREE" value="yes"/>
+      <customization name="CAP_QUERY_HAVING_UNSUPPORTED" value="no"/>
+      <customization name="CAP_QUERY_INCLUDE_GROUP_BY_COLUMNS_IN_SELECT" value="no"/>
+      <customization name="CAP_QUERY_JOIN_PUSH_DOWN_CONDITION_EXPRESSIONS" value="no"/>
```
mattChiaravalloti, line 33: Another semi-related capability, the `CAP_QUERY_JOIN_ACROSS_SCHEMAS` description seems to describe what we call "cross-database lookup", which is supported by MongoSQL, but it defaults to "no". Did you investigate that capability as part of this ticket already and determine the "no" is the correct value for us? (From memory the expression 'schema' in a tableau context maps to 'database' in a mongodb context. I may have that wrong, in which case lmk and we can ignore this comment.)

mattChiaravalloti (changes requested):
These are mostly looking good! A few more questions but I think these are the last ones I have on capabilities.

connector/manifest.xml
```diff
@@ -10,24 +10,36 @@
...
-      <customization name="CAP_QUERY_SUBQUERIES_WITH_TOP" value="yes"/>
-      <customization name="CAP_SUPPORTS_SPLIT_FROM_LEFT" value="yes"/>
-      <customization name="CAP_SUPPORTS_SPLIT_FROM_RIGHT" value="yes"/>
-      <customization name="CAP_SUPPORTS_UNION" value="yes"/>
-      <customization name="CAP_QUERY_ALLOW_PARTIAL_AGGREGATION" value="no"/>
-      <customization name="CAP_QUERY_TIME_REQUIRES_CAST" value="yes"/>
+      <!-- JDBC -->
+      <customization name="CAP_JDBC_QUERY_CANCEL" value="no"/>
+      <customization name="CAP_JDBC_SUPPRESS_ENUMERATE_SCHEMAS" value="yes"/>
     </customizations>
```
nbagnard, line 45: Primary and foreign keys are not really relevant to MongoSQL. Should we turn CAP_JDBC_METADATA_READ_FOREIGNKEYS and CAP_JDBC_METADATA_READ_PRIMARYKEYS off?

rychipman (approved):
LGTM

nbagnard (approved):
LGTM! :)

connector/manifest.xml
bucaojit: I was basing this off the `schemas` term which I am disabling in the connectionMetadata [here](https://github.com/mongodb/mongo-tableau-connector/blob/a33b5cb2170b42b2fcaa1adaa71a4e8048d09d2e/connector/connectionMetadata.xml#L7).  I haven't found a scenario that tests this as the Tableau connection is based on one database.  
I can set this to `yes` as MongoSQL supports cross-database-lookup.

connector/manifest.xml
bucaojit: Sure, I'll make the change

connector/manifest.xml
bucaojit: Thanks, similar to with `CAP_QUERY_CASE_OUT_NO_BOOL_OPS` I was looking for a test case that matched the description of the capability.  Testing a Case with null query confirms it is supported, I will switch this.

connector/manifest.xml
mattChiaravalloti: Hmmm, maybe we don't need to set it to `yes` if we have `schemas` disabled 🤔  I'm not really sure.

Are you able to try to test a custom query like `SELECT * FROM db1.foo JOIN db2.bar`? or does that not work for various other reasons? We could always leave this omitted for now and investigate it in the future if we run into cross-db lookup problems!

mattChiaravalloti: We've already filed a few follow-up tickets from the PR; this could certainly be another follow-up investigation!

bucaojit: Sounds good, I've filed [SQL-754](https://jira.mongodb.org/browse/SQL-754)

The custom query with the different databases works through the Custom SQL option.  That is similar to running it through the ADL shell.  As for what Tableau changes (in terms of the query they run or options enabled/disabled) with that capability I'm not sure.  This can be part of the investigation ticket.

mattChiaravalloti: > The custom query with the different databases works through the Custom SQL option.

Perfect. That's all I was really concerned about anyway!

mattChiaravalloti (approved):

bucaojit:
evergreen merge

PR #5, opened 2022-03-12
SQL-735: Implementing Tableau dialect definition file

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-735

Implementing Tableau dialect file based on documentation here: https://tableau.github.io/connector-plugin-sdk/docs/dialect

### DIFF
--- a/connector/dialect.tdd
+++ b/connector/dialect.tdd
@@ -125,23 +125,36 @@
       <unagg-formula>%1</unagg-formula>
       <argument type='int' />
     </function>
-    <!-- SQL-787
+    <function group='aggregate' name='SUMD' return-type='real'>
+      <formula>SUM(DISTINCT %1)</formula>
+      <unagg-formula>%1</unagg-formula>
+      <argument type='real' />
+    </function>
+    <function group='aggregate' name='SUMD' return-type='int'>
+      <formula>SUM(DISTINCT %1)</formula>
+      <unagg-formula>%1</unagg-formula>
+      <argument type='int' />
+    </function>
     <function group='aggregate' name='STDEV' return-type='real'>
       <formula>STDDEV_SAMP(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='real' />
     </function>
     <function group='aggregate' name='STDEV' return-type='real'>
       <formula>STDDEV_SAMP(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='int' />
     </function>
     <function group='aggregate' name='STDEVP' return-type='real'>
       <formula>STDDEV_POP(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='real' />
     </function>
     <function group='aggregate' name='STDEVP' return-type='real'>
       <formula>STDDEV_POP(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='int' />
-    </function> -->
+    </function>
     <function group='cast' name='DATETIME' return-type='datetime'>
       <formula>CAST(%1 AS TIMESTAMP)</formula>
       <argument type='str' />
@@ -151,19 +164,19 @@
       <argument type='datetime' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DECIMAL)</formula>
+      <formula>CAST(%1 AS DOUBLE)</formula>
       <argument type='real' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DECIMAL)</formula>
+      <formula>CAST(%1 AS DOUBLE)</formula>
       <argument type='int' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DECIMAL)</formula>
+      <formula>CAST(%1 AS DOUBLE)</formula>
       <argument type='str' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DECIMAL)</formula>
+      <formula>CAST(%1 AS DOUBLE)</formula>
       <argument type='bool' />
     </function>
     <function group='cast' name='INT' return-type='int'>
@@ -664,9 +677,8 @@
     <aggregation value='AGG_AVG'/>
     <aggregation value='AGG_MIN'/>
     <aggregation value='AGG_MAX'/>
-    <!-- SQL-787: investigate failing aggregate functions
     <aggregation value='AGG_STDEV'/>
-    <aggregation value='AGG_STDEVP'/> -->
+    <aggregation value='AGG_STDEVP'/>
     <aggregation value='AGG_YEAR'/>
     <aggregation value='AGG_MONTH'/>
     <aggregation value='AGG_DAY'/>
--- a/connector/dialect.tdd
+++ b/connector/dialect.tdd
@@ -183,26 +183,23 @@
       <argument type='bool' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>(CASE
-        WHEN %1 THEN &apos;true&apos;
-        WHEN NOT %1 THEN &apos;false&apos;
-        ELSE NULL END)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='bool' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS CHAR)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='real' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS CHAR)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='int' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS CHAR)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='str' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS CHAR)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='datetime' />
     </function>
     <function group='date' name='MONTH' return-type='int'>
@@ -647,7 +644,7 @@
       <argument type='str' />
     </function>
     <function group='string' name='FIND' return-type='int'>
-      <formula>POSITION(CAST(%1 AS CHAR) IN CAST(%2 AS CHAR))</formula>
+      <formula>POSITION(CAST(%1 AS STRING) IN CAST(%2 AS STRING))</formula>
       <argument type='str' />
       <argument type='str' />
     </function>
--- a/connector/dialect.tdd
+++ b/connector/dialect.tdd
@@ -51,7 +51,7 @@
       <argument type='datetime' />
     </function>
     <function group='aggregate' name='MAX' return-type='bool'>
-      <formula>BOOL_OR(%1)</formula>
+      <formula>MAX(%1)</formula>
       <unagg-formula>%1</unagg-formula>
       <argument type='bool' />
     </function>
@@ -81,7 +81,7 @@
       <argument type='date' />
     </function>
     <function group='aggregate' name='MIN' return-type='bool'>
-      <formula>BOOL_AND(%1)</formula>
+      <formula>MIN(%1)</formula>
       <unagg-formula>%1</unagg-formula>
       <argument type='bool' />
     </function>
@@ -322,11 +322,6 @@
       <argument type='datetime' />
       <argument type='datetime' />
     </function>
-    <function group='operator' name='&amp;&amp;' return-type='bool'>
-      <formula>(%1 AND %2)</formula>
-      <argument type='bool' />
-      <argument type='bool' />
-    </function>
     <function group='operator' name='*' return-type='real'>
       <formula>(%1 * %2)</formula>
       <argument type='real' />
@@ -351,16 +346,6 @@
       <formula>(%1 || %2)</formula>
       <argument type='str' />
       <argument type='str' />
-    </function>
-    <function group='operator' name='+' return-type='datetime'>
-      <formula>(%1 + %2 * INTERVAL &apos;1 DAY&apos;)</formula>
-      <argument type='datetime' />
-      <argument type='real' />
-    </function>
-    <function group='operator' name='+' return-type='date'>
-      <formula>(%1 + %2 * INTERVAL &apos;1 DAY&apos;)</formula>
-      <argument type='date' />
-      <argument type='int' />
     </function>
     <function group='operator' name='-' return-type='real'>
       <formula>(-%1)</formula>
@@ -483,7 +468,7 @@
       <argument type='str' />
     </function>
     <function group='string' name='FIND' return-type='int'>
-      <formula>POSITION(CAST(%1 AS CHAR),CAST(%2 AS CHARß))</formula>
+      <formula>POSITION(CAST(%1 AS CHAR) IN CAST(%2 AS CHAR))</formula>
       <argument type='str' />
       <argument type='str' />
     </function>
--- a/connector/dialect.tdd
+++ b/connector/dialect.tdd
@@ -155,12 +155,72 @@
       <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='int' />
     </function>
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='str' />
+    </function>
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='real' />
+    </function>
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='int' />
+    </function>
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='bool' />
+    </function>
     <function group='cast' name='DATETIME' return-type='datetime'>
       <formula>CAST(%1 AS TIMESTAMP)</formula>
       <argument type='str' />
     </function>
     <function group='cast' name='DATETIME' return-type='datetime'>
       <formula>CAST(%1 AS TIMESTAMP)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='cast' name='DATETIME' return-type='datetime'>
+      <formula>CAST(%1 AS TIMESTAMP)</formula>
+      <argument type='real' />
+    </function>
+    <function group='cast' name='DATETIME' return-type='datetime'>
+      <formula>CAST(%1 AS TIMESTAMP)</formula>
+      <argument type='int' />
+    </function>
+    <function group='cast' name='DATETIME' return-type='datetime'>
+      <formula>CAST(%1 AS TIMESTAMP)</formula>
+      <argument type='bool' />
+    </function>
+    <function group='cast' name='DECIMAL' return-type='real'>
+      <formula>CAST(%1 AS DECIMAL)</formula>
+      <argument type='str' />
+    </function>
+    <function group='cast' name='DECIMAL' return-type='real'>
+      <formula>CAST(%1 AS DECIMAL)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='cast' name='DECIMAL' return-type='real'>
+      <formula>CAST(%1 AS DECIMAL)</formula>
+      <argument type='real' />
+    </function>
+    <function group='cast' name='DECIMAL' return-type='real'>
+      <formula>CAST(%1 AS DECIMAL)</formula>
+      <argument type='int' />
+    </function>
+    <function group='cast' name='DECIMAL' return-type='real'>
+      <formula>CAST(%1 AS DECIMAL)</formula>
+      <argument type='bool' />
+    </function>
+    <function group='cast' name='FLOAT' return-type='real'>
+      <formula>CAST(%1 AS DOUBLE)</formula>
+      <argument type='str' />
+    </function>
+    <function group='cast' name='FLOAT' return-type='real'>
+      <formula>CAST(%1 AS DOUBLE)</formula>
       <argument type='datetime' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
@@ -173,11 +233,15 @@
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
       <formula>CAST(%1 AS DOUBLE)</formula>
+      <argument type='bool' />
+    </function>
+    <function group='cast' name='INT' return-type='int'>
+      <formula>CAST(%1 AS INT)</formula>
       <argument type='str' />
     </function>
-    <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DOUBLE)</formula>
-      <argument type='bool' />
+    <function group='cast' name='INT' return-type='int'>
+      <formula>CAST(%1 AS INT)</formula>
+      <argument type='datetime' />
     </function>
     <function group='cast' name='INT' return-type='int'>
       <formula>CAST(%1 AS INT)</formula>
@@ -189,15 +253,35 @@
     </function>
     <function group='cast' name='INT' return-type='int'>
       <formula>CAST(%1 AS INT)</formula>
+      <argument type='bool' />
+    </function>
+    <function group='cast' name='LONG' return-type='int'>
+      <formula>CAST(%1 AS LONG)</formula>
       <argument type='str' />
     </function>
-    <function group='cast' name='INT' return-type='int'>
-      <formula>CAST(%1 AS INT)</formula>
+    <function group='cast' name='LONG' return-type='int'>
+      <formula>CAST(%1 AS LONG)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='cast' name='LONG' return-type='int'>
+      <formula>CAST(%1 AS LONG)</formula>
+      <argument type='real' />
+    </function>
+    <function group='cast' name='LONG' return-type='int'>
+      <formula>CAST(%1 AS LONG)</formula>
+      <argument type='int' />
+    </function>
+    <function group='cast' name='LONG' return-type='int'>
+      <formula>CAST(%1 AS LONG)</formula>
       <argument type='bool' />
     </function>
     <function group='cast' name='STR' return-type='str'>
       <formula>CAST(%1 AS STRING)</formula>
-      <argument type='bool' />
+      <argument type='str' />
+    </function>
+    <function group='cast' name='STR' return-type='str'>
+      <formula>CAST(%1 AS STRING)</formula>
+      <argument type='datetime' />
     </function>
     <function group='cast' name='STR' return-type='str'>
       <formula>CAST(%1 AS STRING)</formula>
@@ -209,21 +293,33 @@
     </function>
     <function group='cast' name='STR' return-type='str'>
       <formula>CAST(%1 AS STRING)</formula>
-      <argument type='str' />
+      <argument type='bool' />
     </function>
-    <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS STRING)</formula>
+    <function group='date' name='CURRENT_TIMESTAMP' return-type='datetime'>
+      <formula>CURRENT_TIMESTAMP</formula>
+    </function>
+    <function group='date' name='CURRENT_TIMESTAMP' return-type='datetime'>
+      <formula>CURRENT_TIMESTAMP %1</formula>
+      <argument type='int' />
+    </function>
+    <function group='date' name='DAY' return-type='int'>
+      <formula>EXTRACT(DAY FROM %1)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='date' name='EXTRACT' return-type='int'>
+      <formula>EXTRACT(%1 FROM %2)</formula>
+      <argument type='str' />
       <argument type='datetime' />
     </function>
     <function group='date' name='MONTH' return-type='int'>
-      <formula>CAST(EXTRACT(MONTH FROM %1) AS INTEGER)</formula>
+      <formula>EXTRACT(MONTH FROM %1)</formula>
       <argument type='datetime' />
     </function>
     <function group='date' name='NOW' return-type='datetime'>
       <formula>CURRENT_TIMESTAMP</formula>
     </function>
     <function group='date' name='YEAR' return-type='int'>
-      <formula>CAST(EXTRACT(YEAR FROM %1) AS INTEGER)</formula>
+      <formula>EXTRACT(YEAR FROM %1)</formula>
       <argument type='datetime' />
     </function>
     <function group='logical' name='IFNULL' return-type='bool'>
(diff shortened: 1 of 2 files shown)


connector/dialect.tdd
```diff
@@ -1140,119 +511,13 @@
       <part name='OrderBy' value='ORDER BY %1' />
       <part name='Top' value='LIMIT %1' />
     </format-select>
-    <format-string-literal value='Extended' />  <!-- Postgres specific -->
-    <format-true literal='TRUE' predicate='TRUE' /> <!-- Requires 2020.3 or newer -->
-    <icu-date-token-map>
-      <!-- used by DATEPARSE -->
```
bucaojit, line 1146: Removed the date function related sections as I don't think we support functionality to execute them: DATEPARSE, DATEADD, DATETRUNC, etc.

connector/dialect.tdd
```diff
@@ -1,374 +1,133 @@
...
-    <function group='numeric' name='TAN' return-type='real'>
-      <formula>TAN(%1)</formula>
-      <argument type='real' />
-    </function>
-    <function group='numeric;logical' name='ZN' return-type='real'>
-      <formula>COALESCE(%1, 0)</formula>
+    <function group='aggregate' name='AVG' return-type='real'>
+      <formula>AVG(%1)</formula>
+      <unagg-formula>%1</unagg-formula>
       <argument type='real' />
```
mattChiaravalloti, line 16: does `real` cover the `int` case? `AVG(<int>)` is valid just like `AVG(<real>)`, so do we need to overload here, or is `real` sufficient?

connector/dialect.tdd
```diff
@@ -1,374 +1,133 @@
...
       <argument type='str' />
     </function>
-    <function group='string' name='FIND' return-type='int'>
-      <formula>STRPOS(CAST(%1 AS TEXT),CAST(%2 AS TEXT))</formula>
-      <argument type='str' />
-      <argument type='str' />
+    <function group='aggregate' name='COUNT' return-type='int'>
+      <formula>COUNT(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN 0 ELSE 1 END)</unagg-formula>
+      <argument type='datetime' />
```
mattChiaravalloti, line 41: It looks like we have `COUNT(%1)` only for types `bool`, `real`, `str`, and `datetime`. What about `int`? And what about the `local*` types? (I'm not really sure what those are or if they are relevant to us...). Could this be a candidate for the `any` type, or is that too permissive?

connector/dialect.tdd
```diff
@@ -1,374 +1,133 @@
...
+      <formula>SUM(%1)</formula>
+      <unagg-formula>%1</unagg-formula>
+      <argument type='int' />
     </function>
-    <function group='cast' name='DATE' return-type='date'>
-      <formula>CAST(%1 AS DATE)</formula>
-      <argument type='datetime' />
+    <function group='aggregate' name='STDEV' return-type='real'>
+      <formula>STDDEV_SAMP(%1)</formula>
+      <unagg-formula>NULL</unagg-formula>
```
mattChiaravalloti: What does it mean for this to be `NULL`? The docs say:
> Unaggregated formulas should represent a reasonable way of expressing the aggregate of a single value. For example, for average, it should be the value itself. For variance, it should be 0.

We have the option of omitting this element -- should we do that instead of providing null? What is it that we're trying to represent here?

mattChiaravalloti: This applies to STDEVP as well.

connector/dialect.tdd
```diff
@@ -1,374 +1,133 @@
...
+      <argument type='real' />
```
mattChiaravalloti, line 141: Aren't we able to apply this (and STDEVP) to `int` as well?

connector/dialect.tdd
```diff
@@ -379,61 +138,34 @@
...
       <argument type='datetime' />
     </function>
-    <function group='cast' name='FLOAT' return-type='real'>
+    <function group='cast' name='STR' return-type='str'>
       <formula>(CASE
-	WHEN %1 THEN 1.0
-	WHEN NOT %1 THEN 0.0
-	ELSE NULL END)</formula>
+        WHEN %1 THEN &apos;1&apos;
+        WHEN NOT %1 THEN &apos;0&apos;
```
mattChiaravalloti: Is this literally how tableau will execute the function? If so, then should we change `1` and `0` to `true` and `false` since that's how mongosql stringifies bools?

connector/dialect.tdd
```diff
@@ -903,126 +450,36 @@
       <argument type='datetime' />
       <argument type='datetime' />
     </function>
-    <function group='operator' name='^^' return-type='real'>
-      <formula><![CDATA[(CASE WHEN %1 < 0 AND FLOOR(%2) <> %2 THEN NULL ELSE POWER(%1,%2) END)]]></formula>
-      <argument type='real' />
-      <argument type='real' />
-    </function>
     <function group='operator' name='||' return-type='bool'>
```
mattChiaravalloti: In standard SQL-92 (and in MongoSQL), the `||` operator is string concatenation. In this file, it seems we're saying `||` is logical OR. Does this matter?

connector/dialect.tdd
```diff
@@ -553,164 +285,19 @@
...
-    </function>
-    <function group='aggregate' name='MIN' return-type='real'>
-      <formula>MIN(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
-      <argument type='real' />
-    </function>
-    <function group='aggregate' name='MIN' return-type='int'>
-      <formula>MIN(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
+    <function group='numeric' name='DIV' return-type='int'>
```
mattChiaravalloti, line 361: A question about division and arithmetic operators in general:
- MongoSQL generally supports `<any numeric> <arithmetic operator> <any numeric>`, but this file only specifies arithmetic operations were the operands have the same type (int, int or real, real). Do we need to specify the mixed versions (int, real; real, int)?

connector/dialect.tdd
```diff
@@ -379,61 +138,34 @@
       <formula>CAST(%1 AS TIMESTAMP)</formula>
       <argument type='datetime' />
     </function>
-    <function group='cast' name='FLOAT' return-type='real'>
+    <function group='cast' name='STR' return-type='str'>
```
mattChiaravalloti: I don't see all of the various casts that we say we support in mongosql (i.e. what about casting a string to an int or a real or a bool?) Do we need to enumerate all possible valid casts?

connector/dialect.tdd
```diff
@@ -553,164 +285,19 @@
...
-    </function>
-    <function group='aggregate' name='MAX' return-type='bool'>
-      <formula>BOOL_OR(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
-      <argument type='bool' />
-    </function>
-    <function group='aggregate' name='MAX' return-type='real'>
-      <formula>MAX(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
+    <function group='numeric;logical' name='ZN' return-type='real'>
```
mattChiaravalloti: What is "ZN"?

mattChiaravalloti (changes requested):
In addition to my comment threads below, I also noticed we retained an `IIF` function, which we don't actually support in MongoSQL. Are we keeping that because we are able to specify a formula that captures its semantics?

connector/dialect.tdd
```diff
@@ -1,374 +1,133 @@
...
+      <argument type='datetime' />
     </function>
-    <function group='string' name='RIGHT' return-type='str'>
-      <formula>(CASE WHEN %2 &gt;= 0 THEN SUBSTR(CAST(%1 AS TEXT), (LENGTH(CAST(%1 AS TEXT)) - CAST(FLOOR(%2) AS INTEGER) + 1) ) ELSE NULL END)</formula>
-      <argument type='str' />
-      <argument type='real' />
+    <function group='aggregate' name='MAX' return-type='date'>
+      <formula>MAX(%1)</formula>
+      <unagg-formula>%1</unagg-formula>
+      <argument type='date' />
```
rychipman: what does it mean to have this given that MongoSQL doesn't actually support a date (as in date-without-time) type?

rychipman (changes requested):
just one question beyond the things Matt already pointed out

connector/dialect.tdd
bucaojit: I am able to run an `AVG` aggregate on a column with an integer type with the current dialect.  I can err on the side of caution and add an AVG function that explicitly takes an `int` argument in the dialect.  
In my testing, Tableau still has the column type set to `double` with the with the argument type='int' added:
```
2022-03-15 09:01:05.726 -0700 (,,,,2,16) grpc-default-executor-2 : INFO  com.tableau.connect.service.ProtocolService - Running query
   SELECT AVG("grades"."testid") AS "avg:testid:ok"
   FROM "grades"
   HAVING (COUNT(1) > 0)
2022-03-15 09:01:05.728 -0700 (,,,,2,16) grpc-default-executor-2 : INFO  com.tableau.connect.grpc.GrpcProtocolService - End local request 16 /runQuery.
...
2022-03-15 09:01:05.882 -0700 (,,,,2,16) pool-5-thread-1 : INFO  com.tableausoftware.data.ProtobufJDBCResultSet - BindingInfo{columnName='avg:testid:ok', columnTypeName='double', columnLabel='avg:testid:ok', columnType=8, precision=15, scale=15, columnDisplaySize=15, protobufType=DOUBLE, nullable=nullable}
```

connector/dialect.tdd
bucaojit: Regarding the `any`, the documents say that is an allowable type, but it doesn't pass the XML validation step:
```
Reason: attribute type='any': value must be one of ['bool', 'date', 'datetime', 'localint', 'localreal', 'localstr', 'int', 'real', 'spatial', 'str']
```

bucaojit: Regarding the `local*` types, I haven't found any information on it besides them being a possible argument type.  I'm thinking we leave them out for now.  What do you think?  
It's possible the TDVT tests will shed some light on this and some of these functions once we have it running. 

I will add the `int` argument type.

connector/dialect.tdd
bucaojit: It's a function where the `null` values are converted to zero.  This one I've tested and get the expected output.

connector/dialect.tdd
bucaojit: When I looked into this, I saw that the examples had NULL and that a stdev of a single value is undefined.  Running it through our driver I see we return a 0.0, that seems to be the correct value to provide for unagg-formula.

After trying to run the aggregate through Tableau I am seeing the STDDEV_* functions in the GROUP By clause which isn't supported.  This may need further investigation, perhaps I'll leave out these stddev funcitons.

connector/dialect.tdd
bucaojit: Added the `int` argument types

connector/dialect.tdd
bucaojit: changed to `true`, `false`

connector/dialect.tdd
bucaojit: Okay, I removed it so it won't be confusing that the operator has a different behavior.

connector/dialect.tdd
bucaojit: For the `create calculated field` feature I am able to use the operators with int and real values without specifying the argument types.  I'm not sure where else Tableau will use the functions so I added the different variations of the operands.

connector/dialect.tdd
bucaojit: I added the ones I see we support

connector/dialect.tdd
bucaojit: Thanks for clarifying, I wasn't sure if the date was supported as an argument.  I have removed the functions with return type or argument type as `date`.

connector/dialect.tdd
bucaojit: I created a ticket [SQL-787](https://jira.mongodb.org/browse/SQL-787) to further investigate these STDDEV functions, it fails when the aggregation function is in the group by clause and out of the aggregates these are the only ones that fail in that way.

connector/dialect.tdd
```diff
@@ -380,60 +151,69 @@
...
+      <argument type='datetime' />
+    </function>
+    <function group='date' name='NOW' return-type='datetime'>
+      <formula>CURRENT_TIMESTAMP</formula>
+    </function>
+    <function group='date' name='YEAR' return-type='int'>
+      <formula>CAST(EXTRACT(YEAR FROM %1) AS INTEGER)</formula>
       <argument type='datetime' />
     </function>
     <function group='logical' name='IFNULL' return-type='bool'>
```
nbagnard: I don't see ifnull in MongoSQL syntax, only nullif. Did I miss it?

connector/dialect.tdd
mattChiaravalloti: That's ok the result of `avg` is a `double`! The MongoSQL spec (doc & tests) does not indicate that we use integer arithmetic when the argument is an integer anyway. (Unlike for the binary arithmetic operators `+`, `-`, `*`, and `/` where we have more complex rules for how the operations are performed and what the result types are).

I'm a little surprised we don't **need** to have an explicit `AVG(<int>)` since it seems the dialect file requires overloading for every supported type for every function, but maybe `real` is a superset of `int`? 🤷‍♀️

connector/dialect.tdd
mattChiaravalloti: > Regarding the `any`, the documents say that is an allowable type, but it doesn't pass the XML validation step:
> 
> ```
> Reason: attribute type='any': value must be one of ['bool', 'date', 'datetime', 'localint', 'localreal', 'localstr', 'int', 'real', 'spatial', 'str']
> ```

Wow that is unfortunate 👎 

For now, we can leave out the `local` types since we have no information about them. Thanks for adding `int`!

connector/dialect.tdd
mattChiaravalloti: ok thanks for filing the investigation ticket!

connector/dialect.tdd
mattChiaravalloti: So in general, does tableau use the `formula` in SQL queries as opposed to the actual function?

If that is the case, I wonder if we should instead be using our `CAST` function instead of this `CASE` formula. `CASE` statements lead to pretty big translations, whereas a `CAST` is smaller and more straightforward.

connector/dialect.tdd
mattChiaravalloti: Is this a function that tableau requires us to support? Is there a list of default functions they expect?

bucaojit: Here's a list of all the functions supported by Tableau: https://help.tableau.com/current/pro/desktop/en-us/functions_all_alphabetical.htm#Z
I was using these dialects for examples:
https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/Annotated.tdd
https://github.com/tableau/connector-plugin-sdk/tree/master/samples/plugins/postgres_jdbc

If the function is not in the dialect file, then the user will get a message that the function is not available when trying to use it.  It's not required to have the connector pass the package validation.  My thought is that we would want to have more functions implemented if it can be done through the formula field, I can remove them if that is preferred.

connector/dialect.tdd
bucaojit: Ah yes, it uses what is in `formula` for that part of the query, for example here's the calculated field input where `Enrolled` is a boolean column:
```
STR([Enrolled])
```
The statement that Tableau runs is:
```
   SELECT (CASE
           WHEN "class"."enrolled" THEN 'true'
           WHEN NOT "class"."enrolled" THEN 'false'
           ELSE NULL END) AS "Calculation_4639622423314558980"
   FROM "class"
```

mattChiaravalloti: Ok so do we think that's better than `SELECT CAST("class"."enrolled" AS STRING) FROM "class"`? (Or even more succinctly as `SELECT "class"."enrolled"::STRING FROM "class"`? I guess ultimately what I'm asking is, why are we not using the MongoSQL `CAST` function in the formula here?

connector/dialect.tdd
mattChiaravalloti: This is a good design question. We should discuss this at stand-up tomorrow or at doc review on friday!

connector/dialect.tdd
bucaojit: Oh, I see what you mean.  Yes that is more straightforward to use the CAST function here, I made the change.  Also changed the existing as CHAR to as STRING as I see that's how it gets rewritten.

connector/dialect.tdd
bucaojit: I'll bring this up at standup, there are some functions that can be implemented in the `formula` such as `IFNULL` and `ZN` even if the actual function itself is not implemented.  The question is whether to do that in order to have more Tableau functions available or wait until it is implemented in mongosql.

connector/dialect.tdd
```diff
@@ -1,376 +1,147 @@
...
+    <function group='aggregate' name='AVG' return-type='real'>
+      <formula>AVG(%1)</formula>
+      <unagg-formula>%1</unagg-formula>
       <argument type='real' />
     </function>
-    <function group='numeric;logical' name='ZN' return-type='int'>
-      <formula>COALESCE(%1, 0)</formula>
+    <function group='aggregate' name='COUNT' return-type='int'>
+      <formula>COUNT(%1)</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN 0 ELSE 1 END)</unagg-formula>
```
mattChiaravalloti, line 20: Is this `<unagg-formula>` required? How is it used by Tableau in practice? The docs don't make it very clear. Is this being used in a query when `COUNT(x)` is present?

connector/dialect.tdd
```diff
@@ -1,376 +1,147 @@
...
-      <argument type='datetime' />
+    <function group='aggregate' name='SUM' return-type='real'>
+      <formula>SUM(%1)</formula>
+      <unagg-formula>%1</unagg-formula>
+      <argument type='real' />
     </function>
-    <function group='date' name='YEAR' return-type='int'>
-      <formula>CAST(EXTRACT(YEAR FROM %1) AS INTEGER)</formula>
-      <argument type='datetime' />
+    <function group='aggregate' name='SUM' return-type='int'>
```
mattChiaravalloti, line 123: We support `SUM(DISTINCT %1)`. Should we add that to this file?

connector/dialect.tdd
```diff
@@ -1,376 +1,147 @@
...
-    <function group='cast' name='DATETIME' return-type='datetime'>
-      <formula>(CAST(&apos;1900-01-01 00:00:00&apos; AS TIMESTAMP) + %1 * INTERVAL &apos;1 DAY&apos;)</formula>
+    <function group='aggregate' name='STDEVP' return-type='real'>
+      <formula>STDDEV_POP(%1)</formula>
       <argument type='real' />
     </function>
+    <function group='aggregate' name='STDEVP' return-type='real'>
+      <formula>STDDEV_POP(%1)</formula>
+      <argument type='int' />
+    </function> -->
```
mattChiaravalloti: What about the other [aggregate functions](https://docs.google.com/document/d/14DnyMSzVNRkoyib8yM62MPJIrkrG1pE0KjRYOWOUuDc/edit#heading=h.pkz5w745vwh7) we support?
- ADD_TO_ARRAY (optional distinct)
- ADD_TO_SET
- FIRST
- LAST (optional distinct)
- MERGE_OBJECTS

Should we add these to this file or no because they aren't part of the tableau list of functions?

connector/dialect.tdd
```diff
@@ -380,60 +151,66 @@
       <argument type='datetime' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>(CASE
-	WHEN %1 THEN 1.0
-	WHEN NOT %1 THEN 0.0
-	ELSE NULL END)</formula>
-      <argument type='bool' />
+      <formula>CAST(%1 AS DECIMAL)</formula>
```
mattChiaravalloti, line 207: I see this function is named "FLOAT" but we're casting the argument as a `DECIMAL`. Is that the appropriate type mapping? I thought "FLOAT" more closely maps to the MongoSQL `DOUBLE` type. MongoSQL does support casting to `DECIMAL`, I'm just surprised by the name.

connector/dialect.tdd
```diff
@@ -380,60 +151,66 @@
...
+      <argument type='real' />
     </function>
     <function group='cast' name='FLOAT' return-type='real'>
-      <formula>CAST(%1 AS DOUBLE PRECISION)</formula>
-      <argument type='real' />
+      <formula>CAST(%1 AS DECIMAL)</formula>
```
mattChiaravalloti, line 211: In a previous comment, I noted that "FLOAT" should probably map to the MongoSQL `DOUBLE` type instead of `DECIMAL`. Regardless, for casting to `DECIMAL` _or_ `DOUBLE`, we actually support the argument being a datetime since we follow mongodb's `$convert` semantics which do support this for [casting to decimal](https://docs.mongodb.com/manual/reference/operator/aggregation/convert/#converting-to-a-decimal) and for [casting to double](https://docs.mongodb.com/manual/reference/operator/aggregation/convert/#converting-to-a-double). So we should add a case for when the argument type is `datetime`.

connector/dialect.tdd
mattChiaravalloti: There are still several missing cast functions missing:
- all types to `BOOL`
- all types to `DOUBLE` (kinda discussed in previous comments)
- We also support casting to ARRAY, OBJECT, LONG, and OBJECTID, but I'm not sure how we'd represent those in this file since there are not clean `return-type`s for those.

[This section](https://docs.google.com/document/d/14DnyMSzVNRkoyib8yM62MPJIrkrG1pE0KjRYOWOUuDc/edit#heading=h.vjgnkzhj81tw) of the syntax and semantics doc covers everything!

connector/dialect.tdd
```diff
@@ -380,60 +151,66 @@
...
+      <formula>CAST(%1 AS STRING)</formula>
       <argument type='str' />
     </function>
     <function group='cast' name='STR' return-type='str'>
-      <formula>CAST(%1 AS TEXT)</formula>
+      <formula>CAST(%1 AS STRING)</formula>
+      <argument type='datetime' />
+    </function>
+    <function group='date' name='MONTH' return-type='int'>
+      <formula>CAST(EXTRACT(MONTH FROM %1) AS INTEGER)</formula>
```
```suggestion
      <formula>EXTRACT(MONTH FROM %1)</formula>
```
Why do we need to cast the result to int? Doesn't `EXTRACT` return an int?

connector/dialect.tdd
```diff
@@ -524,192 +296,53 @@
...
-      <argument type='real' />
+    <function group='numeric;logical' name='ZN' return-type='int'>
+      <formula>COALESCE(%1, 0)</formula>
+      <argument type='int' />
     </function>
-    <function group='aggregate' name='SUM' return-type='real'>
-      <formula>SUM(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
+    <function group='numeric' name='DIV' return-type='int'>
+      <formula>CASE WHEN %2 = 0 THEN NULL ELSE ( %1 / %2 ) END</formula>
```
mattChiaravalloti: Why are we using this formula when the semantics of MongoSQL's `/` operator already do this for us?

connector/dialect.tdd
```diff
@@ -725,6 +358,21 @@
       <argument type='real' />
       <argument type='real' />
     </function>
+    <function group='operator' name='!=' return-type='bool'>
+      <formula>(%1 &lt;&gt; %2)</formula>
```
mattChiaravalloti: I notice parentheses around a lot of the formulas -- are they strictly necessary? Is this valid:
```
<formular>%1 &lt;&gt; %2</formula>
```
?

connector/dialect.tdd
```diff
@@ -770,63 +413,88 @@
...
+      <formula>(%1 + %2)</formula>
+      <argument type='int' />
+      <argument type='real' />
+    </function>
+    <function group='operator' name='+' return-type='real'>
+      <formula>(%1 + %2)</formula>
+      <argument type='real' />
+      <argument type='int' />
+    </function>
     <function group='operator' name='+' return-type='str'>
```
mattChiaravalloti: From a usability perspective, this concerns me. In the core MongoSQL syntax, `+` has one meaning: arithmetic addition. It is invalid to use it for string concatenation. It looks like here we're saying Tableau users will be able to do string concatenation via the `+` operator. If users get used to that and then move to other tools (or possibly even just write custom SQL in tableau) they may get confused why suddenly `+` doesn't work for string concatenation. We should discuss as a team if we want to expose this operator this way.

connector/dialect.tdd
```diff
@@ -770,63 +413,88 @@
...
-    <function group='operator' name='+' return-type='date'>
-      <formula>(%1 + %2 * INTERVAL &apos;1 DAY&apos;)</formula>
-      <argument type='date' />
-      <argument type='int' />
-    </function>
     <function group='operator' name='-' return-type='real'>
       <formula>(-%1)</formula>
       <argument type='real' />
     </function>
+    <function group='operator' name='-' return-type='int'>
```
mattChiaravalloti, line 433: Good catch. Do we have unary addition anywhere in this file? If not, can we add it?

connector/dialect.tdd
```diff
@@ -770,63 +413,88 @@
...
       <formula>(%1 - %2)</formula>
-      <argument type='int' />
+      <argument type='real' />
       <argument type='int' />
     </function>
-    <function group='operator' name='-' return-type='datetime'>
-      <formula>(%1 - %2 * INTERVAL &apos;1 DAY&apos;)</formula>
-      <argument type='datetime' />
+    <function group='operator' name='/' return-type='real'>
+      <formula>(CASE WHEN %2 = 0 THEN NULL ELSE CAST(%1 AS DOUBLE PRECISION) / %2 END)</formula>
```
mattChiaravalloti: Why are we casting `%1` as a `DOUBLE`?

mattChiaravalloti (changes requested):

connector/dialect.tdd
bucaojit: I tested adding a new function such as `ADD_TO_ARRAY()` and found that my original understanding of the description was incorrect, they give a full list of [supported functions](https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/FullFunctionMap.tdd), but we can add our own custom functions and it will show up in their list of available functions in Tableau.  The difference is it doesn't have a description or helper message like their supported functions do. 
I will add our functions and the different casts, currently there is only what was in the supported functions list.

connector/dialect.tdd
bucaojit: After more testing, I found that the `<unagg-formula>` affects the GROUP BY clause in that if it's not set then the GROUP BY will have the aggregate function in it, which is why STDDEV was failing.  How it affects the GROUP BY clause when it IS set is not clear as so far it hasn't changed based on what I have in the unagg-formula.  

The description of the unagg-formula it `should represent a reasonable way of expressing the aggregate of a single value` which is correct in this case since count(NULL) is 0 and COUNT(N) is 1.

connector/dialect.tdd
bucaojit: I added it as `SUMD`

connector/dialect.tdd
bucaojit: I added some of our functions such as `NULLIF`, `SUBSTRING` as they have supported argument types.  We can discuss the aggregate functions that you mentioned, I'm not sure if they can be represented in the dialect.

connector/dialect.tdd
bucaojit: I added a `DECIMAL` cast function and changed `FLOAT` to cast to `DOUBLE`.

connector/dialect.tdd
bucaojit: I removed the cast to int

connector/dialect.tdd
bucaojit: Updated the formula to use `/` only

connector/dialect.tdd
bucaojit: I tested it and the functions I ran succeeded without parentheses.  I removed them.

connector/dialect.tdd
bucaojit: Sounds good, I've removed it.

connector/dialect.tdd
bucaojit: Sure, added.

connector/dialect.tdd
bucaojit: Removed the cast

connector/dialect.tdd
```diff
@@ -1033,104 +826,18 @@
...
-    <aggregation value='AGG_QTR'/>
     <aggregation value='AGG_MONTH'/>
     <aggregation value='AGG_DAY'/>
-    <aggregation value='AGG_WEEK'/>
-    <aggregation value='AGG_WEEKDAY'/>
-    <aggregation value='AGG_MONTHYEAR'/>
-    <aggregation value='AGG_MDY'/>
     <aggregation value='AGG_HOUR'/>
     <aggregation value='AGG_MINUTE'/>
     <aggregation value='AGG_SECOND'/>
```
bucaojit, line 763: I haven't been able to successfully test aggregating by datetime.  I will open an investigation ticket to look into this further.

connector/dialect.tdd
```diff
@@ -1,375 +1,179 @@
...
-      <formula>CAST(%1 AS DATE)</formula>
-      <argument type='datetime' />
+    <function group='cast' name='BOOL' return-type='bool'>
+      <formula>CAST(%1 AS BOOL)</formula>
+      <argument type='int' />
     </function>
-    <function group='cast' name='DATETIME' return-type='datetime'>
-      <formula>(CAST(&apos;1900-01-01 00:00:00&apos; AS TIMESTAMP) + %1 * INTERVAL &apos;1 DAY&apos;)</formula>
-      <argument type='real' />
+    <function group='cast' name='BOOL' return-type='bool'>
```
nbagnard, line 174: Maybe if you have time, can you test this out: Instead of CAST(%1 AS BOOL), what about just using %1 ? Because the cast is not necessary when cast to the same type.

connector/dialect.tdd
```diff
@@ -379,650 +183,568 @@
...
-      <unagg-formula>%1</unagg-formula>
-      <argument type='real' />
+    <function group='date' name='CURRENT_TIMESTAMP' return-type='datetime'>
+      <formula>CURRENT_TIMESTAMP</formula>
     </function>
-    <function group='aggregate' name='MAX' return-type='int'>
-      <formula>MAX(%1)</formula>
-      <unagg-formula>%1</unagg-formula>
+    <function group='date' name='CURRENT_TIMESTAMP' return-type='datetime'>
+      <formula>CURRENT_TIMESTAMP %1</formula>
```
nbagnard: If I understand this correctly, this is from specifying the precision, right?
From the syntax doc, it should be CURRENT_TIMESTAMP(%1). If CURRENT_TIMESTAMP %1 is also a valid MongoSQL syntax, then you can keep it as is of course (I personally prefer to see the parenthesis because it means less possibly error prone generated queries).

connector/dialect.tdd
```diff
@@ -379,650 +183,568 @@
...
+    <function group='operator' name='+' return-type='int'>
+      <formula>+%1</formula>
       <argument type='int' />
     </function>
     <function group='operator' name='+' return-type='real'>
-      <formula>(%1 + %2)</formula>
+      <formula>+%1</formula>
+      <argument type='real' />
+    </function>
+    <function group='operator' name='+' return-type='real'>
```
nbagnard, line 409: Did you try a mix of arithmetic operators in the same query? Does Tableau add () around them? 
If not we need to add () them in the formulas, because _A + B x C_ does not equal _(A + B) x C_ and this could be a problem.

nbagnard (commented):
I only have a couple of small questions left.
My main worry is around parenthesis or lack of in this case for the arithmetic formulas.
Other than, it looks great! Thank you for figuring out and testing all of this :)

connector/dialect.tdd
bucaojit: Good catch, I'll add the parenthesis back for CURRENT_TIMESTAMP.  It must have been part of my pass of removing parenthesis.

connector/dialect.tdd
bucaojit: Decided in standup today to re-add parentheses to the formulas.  I didn't test the mix of operators previously.  
We also decided we will run TDVT in the future when it is setup to check for correctness of the dialect file.

connector/dialect.tdd
bucaojit: I removed the cast where the argument type matched the type being casted to, STR, INT, FLOAT, DATETIME, BOOL. 
I ran a quick test for `BOOL` and that returned the boolean type as expected

connector/manifest.xml
```diff
@@ -42,6 +42,7 @@
       <customization name="CAP_JDBC_METADATA_READ_PRIMARYKEYS" value="no"/>
       <customization name="CAP_JDBC_QUERY_CANCEL" value="no"/>
       <customization name="CAP_JDBC_SUPPRESS_ENUMERATE_SCHEMAS" value="yes"/>
+      <customization name="CAP_JDBC_USE_ADAPTIVE_FETCH_SIZE" value="no"/>
```
bucaojit, line 45: Modifying this capability here.  To avoid these lines being logged after every query:
```
ERROR com.tableausoftware.jdbc.FetchSizeHelpers - Failed to update fetch size.
java.sql.SQLFeatureNotSupportedException: com.mongodb.jdbc.MongoResultSet.setFetchSize(MongoResultSet.java:535)
```

nbagnard (approved):
LGTM!

connector/dialect.tdd
bucaojit: As discussed at standup, I have removed the functions that are mainly implemented in the formula.  I have created SQL-794 to track the functions supported by Tableau that are not supported by MongoSQL.  We decided it is better to include only the functions that MongoSQL support as these will have been unit tested and would lead to less confusion for functions only in Tableau and not MongoSQL.

connector/dialect.tdd
bucaojit: Added the `BOOL`, `DECIMAL`, `LONG`, and additional casts for `DATETIME`, `FLOAT`.  The ones without a matching return type or arguments were not added.

connector/dialect.tdd
```diff
@@ -1,541 +1,326 @@
...
-	ELSE NULL END)</formula>
-      <argument type='bool' />
+      <formula>CAST(%1 AS INT)</formula>
+      <argument type='str' />
     </function>
     <function group='cast' name='INT' return-type='int'>
-      <formula>CAST(TRUNC(%1) AS BIGINT)</formula>
-      <argument type='real' />
+      <formula>CAST(%1 AS INT)</formula>
+      <argument type='datetime' />
```
mattChiaravalloti: Based on this PR, it seems Tableau's `int` covers 32- and 64-bit ints. In MongoSQL/MongoDB these are separate types, `int` and `long`, respectively. We don't support converting a bson `date` (Tableau `datetime`) to a 32-bit `int` or converting a 32-bit `int` to a bson `date`.

We should remove this version of the `INT` CAST function. I'm not sure we should/could remove the `DATETIME` CAST function whose argument type is `int` since that could be covering the `long` case... Regardless, let's get rid of this version of `INT`.

mattChiaravalloti (approved):
looks good except for one function that needs to be removed 🏁

connector/dialect.tdd
bucaojit: Thanks for that info, I've removed the `INT` CAST of argument type='datetime'.

rychipman (approved):
LGTM

connector/dialect.tdd
bucaojit: Opened [SQL-788](https://jira.mongodb.org/browse/SQL-788)

bucaojit:
evergreen merge

PR #8, opened 2022-04-28
SQL-864: Json datasets for TDVT

nbagnard:


tdvt/README.md
```diff
@@ -0,0 +1,42 @@
...
...a: "Calcs", schema: { "version": 1, "jsonSchema": { bsonType: ['object'], properties: { bool2: { bsonType: ['bool', 'null'] }, date3: { bsonType: ['date', 'null'] }, _id: { bsonType: ['objectId', 'null'] }, num2: { bsonType: ['double', 'null'] }, time1: { bsonType: ['date', 'null'] }, num4: { bsonType: ['double', 'null'] }, int1: { bsonType: ['null', 'int'] }, num1: { bsonType: ['double', 'null'] }, bool0: { bsonType: ['bool', 'null'] }, datetime0: { bsonType: ['date', 'null'] }, date2: { bsonType: ['date', 'null'] }, time0: { bsonType: ['date', 'null'] }, date0: { bsonType: ['null', 'date'] }, int3: { bsonType: ['int', 'null'] }, str0: { bsonType: ['string', 'null'] }, date1: { bsonType: ['date', 'null'] }, int0: { bsonType: ['int', 'null'] }, str3: { bsonType: ['string', 'null'] }, key: { bsonType: ['string', 'null'] }, num0: { bsonType: ['null', 'double'] }, datetime1: { bsonType: ['string', 'null'] }, str1: { bsonType: ['string', 'null'] }, int2: { bsonType: ['int', 'null'] }, zzz: { bsonType: ['string', 'null'] }, bool3: { bsonType: ['null', 'bool'] }, bool1: { bsonType: ['bool', 'null'] }, num3: { bsonType: ['double', 'null'] }, str2: { bsonType: ['string', 'null'] } } } } })
```
mattChiaravalloti: The capital-vs-lowercase collection name issue comes into play here! If we stick with lowercase, we should change this to `"calcs"`.

Also, why do we want to manually set the schema and mark them all as nullable? From the look of the dataset, some are not actually nullable. If tdvt doesn't require each column be marked as nullable, we could use [`db.runCommand({sqlGenerateSchema: 1, setSchemas: true})`](https://www.mongodb.com/docs/atlas/data-federation/query/sql/sqlgenerateschema/).

tdvt/README.md
```diff
@@ -0,0 +1,42 @@
+# TDVT datasets
+
+## Import the datasets
+### Calcs
+```
+mongoimport --uri "{your connection URL}" --db=tdvt --drop --collection calcs --type=json --file={Path to tdvt/datasets directory}/Calcs.json
```
mattChiaravalloti: I also like using all lowercase names for the collections, but don't the tdvt test queries capitalize `Calcs`? when I was debugging a tdvt test for the group by bug the other week, I had to copy the query and then manually edit it to use lowercase "c" instead of capital "C". Maybe the README should suggest using capital "C" so test queries can be copy-pasted more easily?

tdvt/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+### Staples
+```
+mongoimport --uri "{your connection URL}" --db=tdvt --drop --collection calcs --type=json --file={Path to tdvt/datasets directory}/Staples.json
```
mattChiaravalloti:
```suggestion
mongoimport --uri "{your connection URL}" --db=tdvt --drop --collection staples --type=json --file={Path to tdvt/datasets directory}/Staples.json
```
copy-paste!
(also, my suggestion for capitalization applies here too)

tdvt/README.md
```diff
@@ -0,0 +1,42 @@
...
...'string'] }, 'Product In Stock': { bsonType: ['string'] }, 'Supplier Region': { bsonType: ['string'] }, 'Receive Time': { bsonType: ['double'] }, 'Order Date': { bsonType: ['date'] }, 'Prod Type1': { bsonType: ['string'] }, 'Product Base Margin': { bsonType: ['double'] }, 'Order Quarter': { bsonType: ['string'] }, 'Customer Balance': { bsonType: ['double'] }, 'Ship Promo': { bsonType: ['string'] }, '': { bsonType: ['int'] }, 'Item Count': { bsonType: ['int'] }, 'Customer Name': { bsonType: ['string'] }, 'Sales Total': { bsonType: ['double'] }, 'Prod Type2': { bsonType: ['string'] }, Discount: { bsonType: ['double'] }, 'Received Date': { bsonType: ['date'] }, 'Order Year': { bsonType: ['int'] }, 'Customer Segment': { bsonType: ['string'] }, 'Ship Charge': { bsonType: ['decimal'] }, 'Tax Rate': { bsonType: ['double'] }, 'Employee Name': { bsonType: ['string'] }, 'Prod Type3': { bsonType: ['string'] }, 'Order ID': { bsonType: ['string'] }, _id: { bsonType: ['objectId'] }, 'Order Status': { bsonType: ['string'] }, Price: { bsonType: ['decimal'] }, 'Call Center Region': { bsonType: ['string'] }, 'Order Month': { bsonType: ['int'] }, 'Product Container': { bsonType: ['string'] } } } } })
```
mattChiaravalloti: Here, since we are not explicitly making fields nullable, would it be possible to use [`sqlGenerateSchema`](https://www.mongodb.com/docs/atlas/data-federation/query/sql/sqlgenerateschema/) instead? (note the collection capitalization issue for this one, also!)

mattChiaravalloti (changes requested):
Big fan of the step-by-step detail in the README 🙏  There are a couple small things to clean up there, though!

tdvt/README.md
nbagnard: Using the sqlGenerateSchema command does not create schemas matching exactly the tables definition found here:  https://github.com/tableau/connector-plugin-sdk/blob/master/tests/datasets/TestV1/postgres/load_calcs.sql and https://github.com/tableau/connector-plugin-sdk/blob/master/tests/datasets/TestV1/postgres/load_staples.sql

By default, if not specified otherwise, columns are Nullable in SQL that's why all Calcs columns are set nullable.

tdvt/README.md
nbagnard: ah, good catch!
It's because I'm missing a step here. I uploaded the data on a cluster and the collections names are lower-case, but then I used snake-case for the collections in Data Federation.
This means that there is a missing step in the README for the DataFederation config. I'll fix that.

tdvt/README.md
nbagnard: No, it didn't work either. 
Jut so you know, I originally used sqlGenerateSchema but there was small differences and so of it was causing problems to TDVT. This was is not the most simple, but the most foolproof.
I think the differences/problems were around dates, but I honestly don't remember anymore.

tdvt/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+mongoimport --uri "{your connection URL}" --db=tdvt --drop --collection Calcs --type=json --file={Path to tdvt/datasets directory}/Calcs.json
+```
+### Staples
+```
+mongoimport --uri "{your connection URL}" --db=tdvt --drop --collection Staples --type=json --file={Path to tdvt/datasets directory}/Staples.json
+```
+
+## Configure your Data Federation
+Add the following to `databases`
```
bucaojit, line 14: Let's add a bit more detail to this line: such as for the local ADF the configuration change is in the file `./testdata/config/mongodb_local/tenant-config.json`.  For remote ADF ensure the database and collection names match the below configuration.

tdvt/README.md
```diff
@@ -0,0 +1,72 @@
+# TDVT datasets
```
bucaojit, line 1: In this tdvt Readme it would be helpful to add the LogicalQueryFormat we use and the command to run with unsigned taco files:
```
[Datasource]
LogicalQueryFormat = simple
CommandLineOverride = -DDisableVerifyConnectorPluginSignature=true
```

tdvt/README.md
```diff
@@ -0,0 +1,72 @@
...
...'string'] }, 'Product In Stock': { bsonType: ['string'] }, 'Supplier Region': { bsonType: ['string'] }, 'Receive Time': { bsonType: ['double'] }, 'Order Date': { bsonType: ['date'] }, 'Prod Type1': { bsonType: ['string'] }, 'Product Base Margin': { bsonType: ['double'] }, 'Order Quarter': { bsonType: ['string'] }, 'Customer Balance': { bsonType: ['double'] }, 'Ship Promo': { bsonType: ['string'] }, '': { bsonType: ['int'] }, 'Item Count': { bsonType: ['int'] }, 'Customer Name': { bsonType: ['string'] }, 'Sales Total': { bsonType: ['double'] }, 'Prod Type2': { bsonType: ['string'] }, Discount: { bsonType: ['double'] }, 'Received Date': { bsonType: ['date'] }, 'Order Year': { bsonType: ['int'] }, 'Customer Segment': { bsonType: ['string'] }, 'Ship Charge': { bsonType: ['decimal'] }, 'Tax Rate': { bsonType: ['double'] }, 'Employee Name': { bsonType: ['string'] }, 'Prod Type3': { bsonType: ['string'] }, 'Order ID': { bsonType: ['string'] }, _id: { bsonType: ['objectId'] }, 'Order Status': { bsonType: ['string'] }, Price: { bsonType: ['decimal'] }, 'Call Center Region': { bsonType: ['string'] }, 'Order Month': { bsonType: ['int'] }, 'Product Container': { bsonType: ['string'] } } } } })
```
bucaojit:
```suggestion
db.runCommand({ sqlSetSchema: "Staples", schema: { version: NumberLong("1"), jsonSchema: { bsonType: ['object'], properties: { PID: { bsonType: ['int'] }, 'Ship Priority': { bsonType: ['string'] }, 'Total Cycle Time': { bsonType: ['double'] }, 'Employee Salary': { bsonType: ['decimal'] }, 'Ship Mode': { bsonType: ['string'] }, 'Product ID': { bsonType: ['string'] }, 'Gross Profit': { bsonType: ['double'] }, 'Product Name': { bsonType: ['string'] }, 'Fill Time': { bsonType: ['double'] }, 'Market Segment': { bsonType: ['string'] }, 'Ship Handle Cost': { bsonType: ['decimal'] }, 'Supplier Name': { bsonType: ['string'] }, 'Order Quantity': { bsonType: ['double'] }, 'Order Priority': { bsonType: ['string'] }, 'Supplier Balance': { bsonType: ['double'] }, 'Ship Date': { bsonType: ['date'] }, 'Supplier State': { bsonType: ['string'] }, 'Prod Type4': { bsonType: ['string'] }, 'Customer State': { bsonType: ['string'] }, 'Employee Yrs Exp': { bsonType: ['double'] }, 'Employee Dept': { bsonType: ['string'] }, 'Manager Name': { bsonType: ['string'] }, 'Product In Stock': { bsonType: ['string'] }, 'Supplier Region': { bsonType: ['string'] }, 'Receive Time': { bsonType: ['double'] }, 'Order Date': { bsonType: ['date'] }, 'Prod Type1': { bsonType: ['string'] }, 'Product Base Margin': { bsonType: ['double'] }, 'Order Quarter': { bsonType: ['string'] }, 'Customer Balance': { bsonType: ['double'] }, 'Ship Promo': { bsonType: ['string'] }, '': { bsonType: ['int'] }, 'Item Count': { bsonType: ['int'] }, 'Customer Name': { bsonType: ['string'] }, 'Sales Total': { bsonType: ['double'] }, 'Prod Type2': { bsonType: ['string'] }, Discount: { bsonType: ['double'] }, 'Received Date': { bsonType: ['date'] }, 'Order Year': { bsonType: ['int'] }, 'Customer Segment': { bsonType: ['string'] }, 'Ship Charge': { bsonType: ['decimal'] }, 'Tax Rate': { bsonType: ['double'] }, 'Employee Name': { bsonType: ['string'] }, 'Prod Type3': { bsonType: ['string'] }, 'Order ID': { bsonType: ['string'] }, _id: { bsonType: ['objectId'] }, 'Order Status': { bsonType: ['string'] }, Price: { bsonType: ['decimal'] }, 'Call Center Region': { bsonType: ['string'] }, 'Order Month': { bsonType: ['int'] }, 'Product Container': { bsonType: ['string'] } } } } })
```
Changing `Long` to `NumberLong` will allow the command to run correctly with `mongo` and `mongosh`.

bucaojit (changes requested):
Thanks for adding a tdvt readme!

A few suggestions.

tdvt/README.md
mattChiaravalloti: Ah that's too bad they require it to be nullable by default even if the data doesn't contain nulls 😞  but that makes sense, thanks for explaining

tdvt/README.md
mattChiaravalloti: Ooh right, I also forgot about the ADF (is that the correct new acronym for what used to be ADL?) step. That sounds good 👍

tdvt/README.md
mattChiaravalloti: And can we provide some context on what these are?

tdvt/README.md
mattChiaravalloti: Do we even need to wrap it in `NumberLong`? The `sqlSetSchema` command above for `Calcs` just does `version: 1` without a wrapper at all

mattChiaravalloti (changes requested):
almost there! I'll check back one more time to see what the new section Oliver proposed looks like.

tdvt/README.md
bucaojit: Oh that's a good point 😄.  This does work with `version: 1`, no need to use `NumberLong` as I suggested.

terakilobyte:
evergreen retry

PR #10, opened 2022-05-09
SQL-817: Implement VAR and VARP as formulas

nbagnard:


connector/dialect.tdd
```diff
@@ -145,6 +145,16 @@
       <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='int' />
     </function>
+    <function group='aggregate' name='VAR' return-type='real'>
+      <formula>(STDDEV_SAMP(%1) * STDDEV_SAMP(%1))()</formula>
```
```suggestion
      <formula>(STDDEV_SAMP(%1) * STDDEV_SAMP(%1))</formula>
```
Are those `()` at the end intentional or a typo?

connector/dialect.tdd
```diff
@@ -145,6 +145,16 @@
...
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0.0 END)</unagg-formula>
+      <argument type='real' />
+    </function>
+    <function group='aggregate' name='VARP' return-type='real'>
+      <formula>((STDDEV_POP(%1) * STDDEV_POP(%1)))</formula>
```
```suggestion
      <formula>(STDDEV_POP(%1) * STDDEV_POP(%1))</formula>
```
Is one set of parentheses surrounding the formula enough or do we need 2?

mattChiaravalloti (approved):
Looks good 👍 two small question-suggestions! I assume they can either be accepted or there is a valid reason for them, so either way -- no need to re-tag me!

connector/dialect.tdd
nbagnard: No, both are typos. 
I don't know how I did end up with these here 🤪
Fixed now, thank you for spotting them right away!

connector/dialect.tdd
```diff
@@ -145,6 +145,16 @@
       <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0 END)</unagg-formula>
       <argument type='int' />
     </function>
+    <function group='aggregate' name='VAR' return-type='real'>
+      <formula>(STDDEV_SAMP(%1) * STDDEV_SAMP(%1))</formula>
+      <unagg-formula>(CASE WHEN %1 IS NULL THEN NULL ELSE 0.0 END)</unagg-formula>
```
bucaojit: I looked into this one and the unagg-formula should be NULL as the sample of a single value returns null.  There's also the existing STDEV function that can be switched to NULL on lines 130 and 135 if you'd like to make the change. 
https://www.mongodb.com/docs/manual/reference/operator/aggregation/stdDevSamp/#single-value

```suggestion
      <unagg-formula>NULL</unagg-formula>
```

bucaojit (changes requested):
Great work finding that the functions can be implemented in the formulas.  

One comment about the unagg-formula.

connector/dialect.tdd
nbagnard: Yes, you are right.
I originally had NULL, but then I found this in [Tableau docs](https://tableau.github.io/connector-plugin-sdk/docs/dialect)   : 
> unagg-formula element The unagg-formula (unaggregated formula) element is an optional child of function elements. It should be specified only if the function is part of the aggregate group. Unaggregated formulas should represent a reasonable way of expressing the aggregate of a single value. For example, for average, it should be the value itself. **For variance, it should be 0.**
Looks like NULL makes more sense.
Changed it to NULL. Thanks for pointing this out!

bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #11, opened 2022-05-10
SQL-777: Implementing tag-triggered release upload to public S3

bucaojit:
Added instructions to `RELEASE.md`.

Tagged commits will be uploaded to the `translators-connectors-releases` bucket, non-tagged will be uploaded to `mciuploads`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,6 @@
 This repository contains a custom Tableau connector for MongoDB Atlas
 SQL, wrapping our JDBC driver (see
 [mongodb/mongo-jdbc-driver](/mongodb/mongo-jdbc-driver)).
+
+The releases are available for download at the following location.  Replace `<plugin-version>` with plugin version:  
+https://translators-connectors-releases.s3.us-east-1.amazonaws.com/mongo-tableau-connector/mongodb-jdbc-<plugin-version>.taco
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,74 @@
+# Versioning  
+Versions will follow the [semantic versioning](https://semver.org/) system.  
+The following guidelines will be used to determine when each version component will be updated:
+- **major**: backwards-breaking changes 
+- **minor**: functionality added in a backwards compatible manner
+- **patch**: backwards compatible bug fixes
+
+# Release Process
+### Pre-Release Tasks
+
+#### Start Release Ticket
+Move the JIRA ticket for the release to the "In Progress" state.
+Ensure that its fixVersion matches the version being released.
+
+#### Complete the Release in JIRA
+Go to the [SQL releases page](https://jira.mongodb.org/projects/SQL?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=unreleased), and ensure that all the tickets in the fixVersion to be released are closed.
+Ensure that all the tickets have the correct type. Take this opportunity to edit ticket titles if they can be made more descriptive.
+The ticket titles will be published in the changelog.
+
+If you are releasing a patch version but a ticket needs a minor bump, update the fixVersion to be a minor version bump.
+If you are releasing a patch or minor version but a ticket needs a major bump, stop the release process immediately.
+
+The only uncompleted ticket in the release should be the release ticket.
+If there are any remaining tickets that will not be included in this release, remove the fixVersion and assign them a new one if appropriate.
+
+Close the release on JIRA, adding the current date (you may need to ask the SQL project manager to do this).
+
+#### Ensure Evergreen Passing
+Ensure that the build you are releasing is passing the tests on the evergreen waterfall.
+
+### Release Tasks
+
+#### Ensure master up to date
+Ensure you have the `master` branch checked out, and that you have pulled the latest commit from `mongodb/mongo-tableau-connector`.
+
+#### Ensure plugin-version is correct
+Ensure that the `plugin-version` in [manifest.xml](https://github.com/mongodb/mongo-tableau-connector/blob/master/connector/manifest.xml) 
+is updated with the correct version.
+
+#### Create the tag and push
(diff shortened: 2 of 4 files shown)


RELEASE.md
```diff
@@ -0,0 +1,74 @@
...
+#### Ensure Evergreen Passing
+Ensure that the build you are releasing is passing the tests on the evergreen waterfall.
+
+### Release Tasks
+
+#### Ensure master up to date
+Ensure you have the `master` branch checked out, and that you have pulled the latest commit from `mongodb/mongo-tableau-connector`.
+
+#### Ensure plugin-version is correct
+Ensure that the `plugin-version` in [manifest.xml](https://github.com/mongodb/mongo-tableau-connector/blob/master/connector/manifest.xml)
```
nbagnard, line 37: Maybe I missed something, but I don't see this in the evergreen script either.
The idea is to have plugin-version={placeholder} in the repo and have the evergreen script automatically update it to the correct tag version before building.

nbagnard (changes requested):
Just one thing regarding the plugin version in the manifest. We can discuss if what I wrote was not clear or if it does not make sense for any reason.
Otherwise I think everything looks good 👍

RELEASE.md
bucaojit: Oh, I see.  I had the idea that we would manually update the version in the manifest.xml.  
How about for {placeholder} I put '0.0.0' so that the taco build will pass the xml validation.  And then if tag-triggered it will find/replace the value with the tag?  The non-tag-triggered will keep the placeholder '0.0.0'.

nbagnard: Yes, sounds good to me. 0.0.0 seems very reasonable as a place holder 👍 
How uptight is Tableau about the version format, do you know?
For non tag triggered, to differentiate them, maybe we could see how to add the git version after the patch number? I'll create an investigation ticket for this since it does not apply to this JIRA itself, but it's good to keep track of it.

evergreen.yml
```diff
@@ -137,6 +172,18 @@ functions:
         permissions: public-read
         display_name: "Tableau Connector Packaging Logfile"
 
+  "upload release":
+    - command: s3.put
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: mongo-tableau-connector/mongodb_jdbc-signed.taco
+        remote_file: mongo-tableau-connector/mongodb-jdbc-${MONGO_TABLEAU_CONNECTOR_VERSION}.taco
```
nbagnard, line 181: Maybe we should check if the name works for Alexi. I checked the design doc and I did not specify the name of the taco file, so she might have missed that too.

RELEASE.md
```diff
@@ -0,0 +1,74 @@
...
+priority.
+
+### Post-Release Tasks
+
+#### Wait for evergreen
+Wait for the evergreen version to finish, and ensure that the release task completes successfully.
+
+#### Verify release artifacts
+Check that the released TACO file is available at the URL:
+- https://translators-connectors-releases.s3.us-east-1.amazonaws.com/mongo-tableau-connector/mongodb-jdbc-<plugin-version>.taco
```
nbagnard, line 63: Can you also add the URL to the README as well so that users who come from the repo have access to the released taco? I tried to describe this in the "Distribution" section of the spec, but I realize that it is a little succinct. I think for now, just a line saying something around the line "The releases are available for download here : https://translators-connectors-releases.s3.us-east-1.amazonaws.com/mongo-tableau-connector/mongodb-jdbc-<plugin-version>.taco"

nbagnard (changes requested):

RELEASE.md
bucaojit: I experimented earlier with setting it to something like '0.0.0-78b19e4'.  It doesn't work as the validation check seems to want only digits in the version.  It checks against this pattern for validation: 
`['(\\d+)((\\.)(\\d+))*((-)(\\d+))*']`

evergreen.yml
bucaojit: Sounds good. Checking with Alexi.

RELEASE.md
bucaojit: Good catch, I added the URL to the README.

README.md
```diff
@@ -3,3 +3,6 @@
 This repository contains a custom Tableau connector for MongoDB Atlas
 SQL, wrapping our JDBC driver (see
 [mongodb/mongo-jdbc-driver](/mongodb/mongo-jdbc-driver)).
+
+The releases are available for download at the following location.  Replace `<plugin-version>` with plugin version:  
+https://translators-connectors-releases.s3.us-east-1.amazonaws.com/mongo-tableau-connector/mongodb-jdbc-<plugin-version>.taco
```
nbagnard, line 8: What about providing the bucket link only and omit the file name containing the plugin-version so that they can then choose which one they want?

RELEASE.md
nbagnard: ok, cool, it's nice you already checked. I'll add this as a comment on the related JIRA then. Maybe we can simply add the version as a suffix for the taco file name.

evergreen.yml
nbagnard: Great, thank you 😊

nbagnard (approved):
LGTM! I just have a small comment/question regarding the URL provided in the README that I leave to your best judgment.

README.md
bucaojit: Discussed offline and decided to leave as-is since the bucket link only would lead to an Access Denied page.

rychipman (approved):
code changes LGTM! once we merge this PR (but before closing the ticket), let's try to tag a release to make sure all of this works as expected with the tag-triggered versions

bucaojit:
evergreen merge

PR #12, opened 2022-05-10
SQL_841: Add string functions formulas

nbagnard:


connector/dialect.tdd
```diff
@@ -625,6 +625,87 @@
...
+     <argument type='real' />
+    </function>
+    <function group='string' name='MID' return-type='str'>
+     <formula>(CASE WHEN %3 &lt;= 0 THEN (CASE WHEN %1 IS NULL THEN NULL ELSE '' END) ELSE SUBSTRING(%1 FROM ((%2)::INT - 1) FOR %3)  END)</formula>
+     <argument type='str' />
+     <argument type='real' />
+     <argument type='int' />
+    </function>
+    <function group='string' name='RIGHT' return-type='str'>
+     <formula>(CASE WHEN %2 &lt;= 0 THEN (CASE WHEN %1 IS NULL THEN NULL ELSE '' END) WHEN %2 &gt;= CHAR_LENGTH(%1) THEN %1 ELSE SUBSTRING(%1 FROM (CHAR_LENGTH(%1) - %2 - 1) FOR %2) END)</formula>
```
bucaojit:
```suggestion
     <formula>(CASE WHEN %2 &lt;= 0 THEN (CASE WHEN %1 IS NULL THEN NULL ELSE '' END) WHEN %2 &gt;= CHAR_LENGTH(%1) THEN %1 ELSE SUBSTRING(%1 FROM (CHAR_LENGTH(%1) - %2) FOR %2) END)</formula>
```
There may be an off by one error with the 'RIGHT' formula.  
For example if I try `SELECT SUBSTRING('calculation' FROM (CHAR_LENGTH('calculation') - 4 - 1) FOR 4)`
I get `atio` and not the expected `tion`.

bucaojit (changes requested):
Nice work finding how to implement the formulas using the existing functions!

I had one comment on the `RIGHT` function

connector/dialect.tdd
nbagnard: You are totally right!
Fixed it now :)

bucaojit (approved):
Thanks for fixing the RIGHT formula. 
Looks good.

varsha804 (approved):
lgtm!

nbagnard:
evergreen merge

PR #13, opened 2022-05-10
SQL-865: Setting default property values for logging parameters

bucaojit:
Setting default values for `LogDir` and `LogLevel` properties in case it is not set in existing and created *.tds and *.twb files.

### DIFF
--- a/connector/connectionProperties.js
+++ b/connector/connectionProperties.js
@@ -6,12 +6,12 @@
     props["dialect"] = "mongosql";
 
     props["LogDir"] = "console";
-    if (attr["v-log-directory"] != null && !attr["v-log-directory"] === "") {
+    if (attr["v-log-directory"] != null && attr["v-log-directory"] !== "") {
         props["LogDir"] = attr["v-log-directory"];
     }
 
     props["LogLevel"] = "OFF";
-    if(attr["v-loglevel"] != null && !attr["v-loglevel"] === "") {
+    if(attr["v-loglevel"] != null && attr["v-loglevel"] !== "") {
         props["LogLevel"] = attr["v-loglevel"];
     }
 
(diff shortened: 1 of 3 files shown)


connector/connectionProperties.js
```diff
@@ -5,13 +5,15 @@
     props["database"] = attr[connectionHelper.attributeDatabase];
     props["dialect"] = "mongosql";
 
-    if (attr["v-log-directory"] == null) {
-        props["LogDir"] = "console";
-    } else {
+    props["LogDir"] = "console";
+    if (attr["v-log-directory"] != null && !attr["v-log-directory"] === "") {
```
nbagnard: I tried out the new taco file and this does not work. I always end up with OFF and Console even when changing the LogLevel to FINER for example.
This works : `if (attr["v-log-directory"] != null && attr["v-log-directory"] !== "")`

connector/connectionProperties.js
```diff
@@ -5,13 +5,15 @@
...
         props["LogDir"] = attr["v-log-directory"];
     }
 
-    props["LogLevel"] = attr["v-loglevel"];
+    props["LogLevel"] = "OFF";
+    if(attr["v-loglevel"] != null && !attr["v-loglevel"] === "") {
```
nbagnard: And similarly, here, you want to use : `if(attr["v-loglevel"] != null && attr["v-loglevel"] !== "")`

nbagnard (changes requested):

connector/connectionProperties.js
bucaojit: Thanks for checking.  I made the suggested change.

connector/connectionProperties.js
bucaojit: Made this change as well.

nbagnard (approved):
LGTM!
No idea why your original code did not work ¯\_(ツ)_/¯
Maybe extra parenthesis around the negated clause would have worked too.

bucaojit:
evergreen merge

PR #14, opened 2022-05-13
SQL-883: Add property to indicate connections made with Tableau

bucaojit:


evergreen.yml
```diff
@@ -113,6 +117,19 @@ functions:
             --notary-url http://notary-service.build.10gen.cc:5000 \
             mongodb_jdbc.taco
 
+  "update plugin-version":
```
bucaojit: I have a similar change in PR 11, I will merge the change based on which is checked in first.

nbagnard: 👍 So that it will update both manifest.xml and connectionProperties.js, correct?

nbagnard (approved):

evergreen.yml
bucaojit: Yes, that's correct.

mattChiaravalloti (approved):

bucaojit:
evergreen merge

PR #17, opened 2022-05-27
SQL-840: Implementing functions LN, LOG, ROUND in dialect definition

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-840
Implementing a subset of numeric functions in the dialect definition file

### DIFF
--- a/connector/dialect.tdd
+++ b/connector/dialect.tdd
@@ -286,6 +286,70 @@
       <argument type='real' />
       <argument type='int' />
     </function>
+    <function group='numeric' name='LN' return-type='real'>
+      <formula>LOG(%1, 2.718281828459046)</formula>
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='LN' return-type='real'>
+    <formula>LOG(%1, 2.718281828459046)</formula>
+      <argument type='real' />
+    </function>
+    <function group='numeric' name='LOG' return-type='real'>
+      <formula>LOG(%1, 10)</formula>
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='LOG' return-type='real'>
+      <formula>LOG(%1, 10)</formula>
+      <argument type='real' />
+    </function>
+    <function group='numeric' name='LOG' return-type='real'>
+      <formula>LOG(%1, %2)</formula>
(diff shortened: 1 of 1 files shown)


connector/dialect.tdd
```diff
@@ -286,6 +286,70 @@
       <argument type='real' />
       <argument type='int' />
     </function>
+    <function group='numeric' name='LN' return-type='real'>
+      <formula>LOG(%1, 2.718281828459046)</formula>
```
mattChiaravalloti, line 290: Similar to `PI`, is this constant "good enough" to give correct results in all cases?

mattChiaravalloti (approved):
Looks good assuming the `LN` formula is sufficient! I guess we'll only really be able to know when we can run the tdvt tests but that requires the group by fix. This is probably at least good enough for now -- we can always address `LN` bugs later if necessary, not sure how popular that function will be anyway 🎩

varsha804 (approved):
Also curious about the `e` constant, but otherwise LGTM!

connector/dialect.tdd
bucaojit: That's a good question, I'll take a look at what the TDVT expects in the test and see if it matches what is returned from mongodb.

bucaojit: I compared the results from mongodb with `$log : [ %1, 2.718281828459046 ]` to the TDVT results and the results from mongodb have more precision whereas the TDVT results seem rounded such as:
LN(3)
mongodb: 1.0986122886681096    TDVT: 1.0986123
LN(2)
mongodb: 0.6931471805599452    TDVT: 0.69314718
LN(15.7)
mongodb: 2.753660712354262      TDVT: 2.7536607

It seems then that the value for `e` produces an accurate result for TDVT.   As for the precision differences, we may have to get the test to fully run to see whether the test framework is rounding the value when doing the comparison checking.

bucaojit:
evergreen merge

connector/dialect.tdd
mattChiaravalloti: Ok that sounds good. Looks like the constant we're using is pretty decent, so hopefully we will be able to just stick with it 🤞

PR #18, opened 2022-05-31
SQL-899: Enable split in Tableau connector

nbagnard:
Works with only native split in the tdd (SPLIT appears in the function list and works as expected), so I did not add the string SPLIT formula.


bucaojit (approved):
Very cool, I tried out the connector and was able to split a string field!

varsha804 (approved):
Nice, LGTM!

nbagnard:
evergreen merge

PR #19, opened 2022-06-01
SQL-815: Add logical functions IFNULL, IIF, ZN and operator OR to Tableau dialect

nbagnard:


connector/dialect.tdd
```diff
@@ -266,6 +266,104 @@
...
+    <function group='logical' name='IFNULL' return-type='real'>
+      <formula>COALESCE(%1, %2)</formula>
+      <argument type='real' />
+      <argument type='real' />
+    </function>
+    <function group='logical' name='IFNULL' return-type='int'>
+      <formula>COALESCE(%1, %2)</formula>
+      <argument type='int' />
+      <argument type='int' />
+    </function>
```
```suggestion
    <function group='logical' name='IFNULL' return-type='int'>
      <formula>COALESCE(%1, %2)</formula>
      <argument type='int' />
      <argument type='int' />
    </function>
    <function group='logical' name='IFNULL' return-type='real'>
      <formula>COALESCE(%1, %2)</formula>
      <argument type='real' />
      <argument type='real' />
    </function>
```
[**super** nit] For the two forms of `IIF` we order the types as `bool`, `int`, `real`, ... . Can we swap `real` and `int` here to match that order? _Super_ nitpicky so it's okay not to do this, but it could be helpful when skimming this file to see things in the same order consistently.

mattChiaravalloti (approved):
Looks good! 🥳  My one suggestion is up to you to accept or ignore.

Also, can you remind me what we concluded about polymorphic return types in tableau? They cannot be represented, right? That's why we don't have all combinations of argument types for these functions? For example, `IFNULL(str, int)` or `IIF(bool, datetime, str)`, etc.

nbagnard:
> Also, can you remind me what we concluded about polymorphic return types in tableau? They cannot be represented, right? That's why we don't have all combinations of argument types for these functions? For example, `IFNULL(str, int)` or `IIF(bool, datetime, str)`, etc.

So, from what is in the tdd and what I remember, we decided not to add all the combinations because it would make the tdd very, very verbose.
In TDVT I have only seen this to be a problem with the operators + and ==.
Since MongoSQL is strongly typed, we might want to revisit this if it becomes a pain point for users.

bucaojit (approved):
LGTM!

Regarding the IIF, I see that the argument types here cover the cases in the [full dialect sample](https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/FullFunctionMap.tdd) and the queries in the TDVT have matching types such as `iif(bool0,num0,num1,num2)` and `iif(bool0,date0,date1,date2)`.  I think we can add it later when necessary if a user requests it.

nbagnard:
evergreen merge

PR #21, opened 2022-06-13
SQL-840: Implemented Tableau numeric functions in dialect definition

bucaojit:


connector/dialect.tdd
```diff
@@ -364,6 +364,50 @@
...
+    </function>
+    <function group='numeric' name='ABS' return-type='int'>
+      <formula>ABS(%1)</formula>
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='CEILING' return-type='real'>
+      <formula>CEIL(%1)</formula>
+      <argument type='real' />
+    </function>
+    <function group='numeric' name='CEILING' return-type='int'>
```
nbagnard, line 379: Is there a specific reason to add this combination (return int, param real) and not the other one (return real, param int)?

connector/dialect.tdd
```diff
@@ -384,6 +428,26 @@
...
     </function>
+    <function group='numeric' name='EXP' return-type='real'>
+      <formula>POW(2.718281828459046, %1)</formula>
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='EXP' return-type='real'>
+      <formula>POW(2.718281828459046, %1)</formula>
+      <argument type='real' />
+    </function>
+    <function group='numeric' name='FLOOR' return-type='real'>
```
nbagnard, line 435: I'm not too sure about the FLOOR formula. $floor returns an int. I think MongoSQL floor does as well from the spec "The FLOOR ( number ) scalar function returns a number to the nearest whole number of equal or lesser value". Why supporting FLOOR returning a real? And if we do, shouldn't the cast to NUMERIC be here instead of a cast INT as done below?

connector/dialect.tdd
```diff
@@ -420,6 +484,62 @@
...
+    </function>
+    <function group='numeric' name='PI' return-type='real'>
+      <formula>3.141592653589793</formula>
+    </function>
+    <function group='numeric' name='POWER' return-type='int'>
+      <formula>POW(%1,%2)</formula>
+      <argument type='int' />
+      <argument type='int' />
+    </function>
+    <function group='numeric' name='POWER' return-type='real'>
```
nbagnard, line 507: Similarly to my question for FLOOR, is there a specific reason to add this combination?

nbagnard (changes requested):

connector/dialect.tdd
bucaojit: I was following what was in the [function list](https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/FullFunctionMap.tdd).   That makes sense though that we would not reach an `int` return type from a `real` argument.  I'll remove these.  It did not cause any test failures from removing so I think it's good.

connector/dialect.tdd
bucaojit: I removed the function with the `int` arg and `real` return.  It makes sense that it would not be reachable.

As for returning int, from what I see in mongodb the type it returns matches the input type.  Maybe the documentation is not referring to the actual return type?
For example for $floor:
```
foo> db.ceilAndFloor.aggregate([ {$addFields : { "n": {$floor: "$v"}}}])
[
  { _id: 0, v: 7, n: 7 },
  { _id: 1, v: Long("7"), n: Long("7") },
  { _id: 2, v: -7.8, n: -8 }, <- this -8 is a Double type.
  { _id: 3, v: Decimal128("7.8"), n: Decimal128("7") }, <- Decimal output 
```

connector/dialect.tdd
bucaojit: Same reason, to match the example.  After testing and seeing that removing it does not affect the test failure count I have removed the `int` arguments and `real` return.

connector/dialect.tdd
nbagnard: oh, I see. Yes, the docs is not very clear on the type. That's fine then. Thanks for looking into this :)

nbagnard (approved):
LGTM!

varsha804 (approved):
LGTM! Sorry for the late review -- I missed the email about this PR!

bucaojit:
evergreen merge

PR #25, opened 2022-09-02
SQL-919: Add formula to concatenate strings when using a + operator

nbagnard:


connector/dialect.tdd
```diff
@@ -638,6 +638,11 @@
       <argument type='int' />
       <argument type='real' />
     </function>
+    <function group='operator' name='+' return-type='str'>
+      <formula>($1 || $2)</formula>
```
bucaojit:
```suggestion
      <formula>(%1 || %2)</formula>
```
I ran this through TDVT and was getting an `invalid token` error, looks like we want the `%` to indicate the substitution.

bucaojit (changes requested):

connector/dialect.tdd
nbagnard: Yes, you are right! Not sure how I got the '$' in here...
Thank you for letting me know!

bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #28, opened 2022-09-02
SQL-1009: Update POW formula and ^^ operator with int arguments to return a real

nbagnard:


bucaojit (approved):
Changes look good.  
Something to note, one of the Required tests using `^^` is failing: num0 ^ num1
```
SELECT POW("Calcs"."num0", "Calcs"."num1") AS "TEMP(Test)(637953353)(0)"
FROM "Calcs"
GROUP BY "TEMP(Test)(637953353)(0)"
```
The expected and actual results look close but the actual has two null results instead of one.  It could be because we return `NaN` and `null` in the results and it's counting it as two separate nulls?  Probably something we can document if there isn't a simple solution.

Expected:
```
%null%
0.0
83910.402
1.5031456e+09
2.9512092e+09
4.9647639e+11
```
Actual:
```
%null%
%null%
0.0
83910.402
1.5031456e+09
2.9512092e+09
4.9647639e+11
```

bucaojit:
I see the NaN to Null problem is already addressed in SQL-1013 👍

nbagnard:
evergreen merge

PR #35, opened 2022-09-08
SQL-1013: Convert NaN to Null in Tableau formulas using POW()

nbagnard:


bucaojit (changes requested):
`POW()` is also used in EXP on line 427, let's add this same NaN to Null logic there.

nbagnard:
> `POW()` is also used in EXP on line 427, let's add this same NaN to Null logic there.
Good catch! Thanks a lot, I missed this one.
It's fixed now.

bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #39, opened 2022-10-14
SQL-1017: Add formula for supporting endswith

nbagnard:


bucaojit (approved):
LGTM!

connector/dialect.tdd
```diff
@@ -1054,6 +1054,11 @@
      <argument type='str' />
      <argument type='str' />
     </function>
+    <function group='string' name='ENDSWITH' return-type='bool'>
+      <formula>(CASE WHEN (CHAR_LENGTH(%1) - CHAR_LENGTH(%2)) &lt; 0 THEN false ELSE SUBSTRING(%1 FROM (CHAR_LENGTH(%1) - CHAR_LENGTH(%2)) FOR (CHAR_LENGTH(%2))) = %2 END)</formula>
```
```suggestion
      <formula>(CASE WHEN CHAR_LENGTH(%1) &lt; CHAR_LENGTH(%2) THEN false ELSE SUBSTRING(%1 FROM (CHAR_LENGTH(%1) - CHAR_LENGTH(%2)) FOR (CHAR_LENGTH(%2))) = %2 END)</formula>
```
why not do `char_length(%1) < char_length(%2)` and save us the additional operation?

mattChiaravalloti (approved):
looks good but I think it can be better -- no need to retag me, I'm ok with this staying the same or changing (the difference is likely very small)

connector/dialect.tdd
nbagnard: Seems very reasonable to me 👍

nbagnard:
evergreen merge

PR #46, opened 2022-10-19
SQL-1012: Update division operator to return real and not int

nbagnard:
The change had been made for the function DIV, but the not the `/` operator


bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #48, opened 2022-10-20
SQL-1012: Cast division arguments from int to double

nbagnard:
Otherwise the division result is an int which Tableau cast to real instead of having MongoSQL perform the correct operation.


bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #49, opened 2022-10-20
SQL-1012: Update formula to return an int for DIV(int,int)

nbagnard:
Note that the `/` operator returns a real and not a int to match TDVT expected behavior.


bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #50, opened 2022-10-20
SQL-1077: Update formula to allow cast from string to int for all numeric values

nbagnard:
Update formula to cast string to number first and then to int to avoid Null when casting floats and numeric values


bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #51, opened 2022-10-20
SQL-1079: Update formula to match Tableau expected behavior for negative indexes

nbagnard:
We need to use FLOOR(index) because otherwise when casting to int the value is simply truncated and the result of RIGHT is off by one character.


pmeredit (approved):
LGTM

nbagnard:
evergreen merge

PR #53, opened 2023-04-18
SQL-1357: DownloadCenter json feed

nbagnard:
The json template will be used to create the current download feed which the download center will use to provide links to users.


terakilobyte (commented):
🚀 I verified the sed replacement and the download works for the latest version (1.2.0). 🔥 

I wish we could see how this gets incorporated into the download center. Additionally, I noticed a broken link for the release notes. We also don't have release notes on in the github repo. Is this something that the docs team will fix prior to this going live, or should we change the link and generate some release notes? Either way, not a blocker.

How do we plan to automate additions to the json `versions` array, or will it always be the latest version?

nbagnard:
> Additionally, I noticed a broken link for the release notes.

This was the link 2 weeks ago. It is now `https://www.mongodb.com/docs/atlas/release-notes/sql/`. I'll update that, thanks for catching that

nbagnard:
> How do we plan to automate additions to the json `versions` array, or will it always be the latest version?
It will always be the last version. When a release is triggered, it uses the template to generate the new version of the feed and update it on S3.

mongo-tableau-connector-downloads_template.json
```diff
@@ -0,0 +1,10 @@
+{
+  "manual_link": "https://www.mongodb.com/docs/atlas/data-federation/query/query-with-sql/",
+  "release_notes_link": "https://www.mongodb.com/docs/atlas/data-federation/query/sql/release-notes/",
+  "versions": [
+    {
+      "version": "{RELEASE_VERSION}",
```
bucaojit, line 6: Wanted to double check that `platform` does not need to be specified here?  As in versions -> platform -> download_link.
Not sure how the download center parses it but the 2 example I see in `mongodb-bi-downloads.json` and the SQL-1358 PR so.

evergreen.yml
```diff
@@ -248,3 +274,4 @@ buildvariants:
       - name: compile
       - name: sign
       - name: release
+      - name: downloadcenter-update
```
bucaojit: Super small nit: Would be clearer for me with a dash
```suggestion
      - name: download-center-update
```

evergreen.yml
```diff
@@ -208,6 +208,26 @@ functions:
...
+          echo "-------------------------"
+    - command: s3.put
+      params:
+        aws_key: ${release_aws_key}
+        aws_secret: ${release_aws_secret}
+        local_file: mongo-tableau-connector/mongo-tableau-connector-downloads_template.json
+        remote_file: mongo-tableau-connector/mongo-tableau-connector-downloads.json
+        bucket: translators-connectors-releases
+        permissions: public-read
+        content_type: application/octet-stream
```
bucaojit:
```suggestion
        content_type: application/json
```
I normally see `octet-stream` for our binaries.  This seems more appropriate, and [here's](https://github.com/mongodb/mongo-tools/blob/e62cf8866c5b98c45faf2fbe354703969aa076c6/common.yml#L411) one instance of json being used.

mongo-tableau-connector-downloads_template.json
nbagnard: There is no platform for a taco file, right?
I shared templates of the json files with the team responsible for the download centre integration and did not hear anything from them. I'll make sure to keep an eye on this.

bucaojit: > There is no platform for a taco file, right?  

That's right, it's the same taco file used for Mac/Windows.

> I shared templates of the json files with the team responsible for the download centre integration and did not hear anything from them.

Cool, sounds good then if they've already checked it.

bucaojit (approved):
LGTM

terakilobyte (approved):
LGTM!

evergreen.yml
terakilobyte: Just a comment here. `application/octet-stream` should be used for types that don't have a dedicated encoding according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types). A full list of dedicated content-types can be found [here](https://www.iana.org/assignments/media-types/media-types.xhtml#application).

nbagnard:
evergreen merge

PR #54, opened 2023-06-15
SQL-1450: Update signing tasks to use the new Garasign platform - Tableau connector

EthanHardyMongo:
This ticket covers replacing uses of Notary Service with Garasign. In this ticket, I created a few new evergreen variables: `garasign_jarsigner_image`, `release_tools_container_registry`, `sql_engines_artifactory_username`, `sql_engines_artifactory_auth_token`, `tableau_garasign_username`, and `tableau_garasign_password`.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -127,7 +127,7 @@ functions:
 
           docker run \
             -e GRS_CONFIG_USER1_USERNAME=${garasign_username} \
-            -e GRS_CONFIG_USER1_PASSWORD=${garasign_password} \
+            -e GRS_CONFIG_USER1_PASSWORD=${tableau_garasign_password} \
             --rm \
             -v $(pwd):$(pwd) -w $(pwd) \
             ${garasign_jarsigner_image} \
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -190,11 +190,31 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongo-tableau-connector/mongodb_jdbc-signed.taco
+        local_file: mongo-tableau-connector/mongodb_jdbc.taco
         remote_file: mongo-tableau-connector/artifacts/${version_id}/mongodb_jdbc-signed.taco
         bucket: mciuploads
         permissions: public-read
         content_type: application/octet-stream
+
+  "get signed connector":
+    - command: s3.get
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        remote_file: mongo-tableau-connector/artifacts/${version_id}/mongodb_jdbc-signed.taco
+        bucket: mciuploads
+        local_file: mongo-tableau-connector/mongodb_jdbc-signed.taco
+
+  "verify signed connector":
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -253,6 +273,8 @@ tasks:
       - func: "fetch packaged connector"
       - func: "sign archive"
       - func: "upload signed connector"
+      - func: "get signed connector"
+      - func: "verify signed connector"
```
EthanHardyMongo: I created these two functions in order to verify that the signed `mongo_jdbc.taco` was put in s3 since it was unknown if Garasign created a signed file like the one created by Notary (i.e., `mongodb_jdbc-signed.taco`). I left this in to show how I verified it, but I can get rid of this before merging.

evergreen.yml
```diff
@@ -190,12 +190,32 @@ functions:
       params:
         aws_key: ${aws_key}
```
bucaojit, line 191: Small nit, the display name shows up as the local_file name which is `mongodb_jdbc.taco`.  
You can set it to `mongodb_jdbc-signed.taco`
Here's an example of display_name: 
https://github.com/mongodb/mongo-tableau-connector/blob/70c4add74ee764a4876d71e59c0cc8c2e293089a/evergreen.yml#L174

evergreen.yml
```diff
@@ -190,12 +190,32 @@ functions:
...
+  "verify signed connector":
+    - command: shell.exec
+      type: system
+      params:
+        working_dir: mongo-tableau-connector
+        silent: false
+        script: |
+          ${prepare_shell}
+          
+          jarsigner -verify mongodb_jdbc-signed.taco -verbose -certs -strict
```
bucaojit: Nice work adding the verification step

evergreen.yml
```diff
@@ -190,12 +190,32 @@ functions:
...
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongo-tableau-connector/mongodb_jdbc-signed.taco
+        local_file: mongo-tableau-connector/mongodb_jdbc.taco
         remote_file: mongo-tableau-connector/artifacts/${version_id}/mongodb_jdbc-signed.taco
         bucket: mciuploads
         permissions: public-read
         content_type: application/octet-stream
 
+  "get signed connector":
```
bucaojit: We have a `fetch signed connector` command already that you can use instead of adding this one.

evergreen.yml
bucaojit: You can add the `verify signed connector` after `sign archive` and you won't need to fetch it from s3 because it will then check it before uploading.  An alternative is you can add another step to `sign archive` and call it `sign and verify archive` and do the work there, as in add another  ` - command: shell.exec` step to sign archive.

bucaojit (changes requested):
Great job getting all the moving parts in place to get the garasign working 👍

bucaojit (approved):
LGTM!

nbagnard (approved):
LGTM!
Very good idea to add the verification!

EthanHardyMongo:
evergreen merge

PR #56, opened 2024-02-09
SQL-1910: account for default version in taco file

terakilobyte:
There was a change in the tableau connector plugin sdk where taco files now have a default version attached, `-v0.0.0`. This PR accounts for that in our upload task.

Additionally, to comply with security requirements, the signing task is now silent to avoid logging any activity in evergreen.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -154,7 +154,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongo-tableau-connector/package/mongodb_jdbc.taco
+        local_file: mongo-tableau-connector/package/mongodb_jdbc-v0.0.0.taco
         remote_file: mongo-tableau-connector/artifacts/${version_id}/mongodb_jdbc.taco
         bucket: mciuploads
         permissions: public-read
@@ -205,7 +205,7 @@ functions:
         silent: false
         script: |
           ${prepare_shell}
-          
+
           jarsigner -verify mongodb_jdbc.taco -verbose -certs -strict
 
   "validate connector XML format":
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -120,7 +120,7 @@ functions:
       type: system
       params:
         working_dir: mongo-tableau-connector
-        silent: false
+        silent: true
         script: |
           ${prepare_shell}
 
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -120,7 +120,7 @@ functions:
       type: system
       params:
         working_dir: mongo-tableau-connector
-        silent: false
+        silent: true
```
terakilobyte, line 125: security change

evergreen.yml
```diff
@@ -154,7 +154,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongo-tableau-connector/package/mongodb_jdbc.taco
+        local_file: mongo-tableau-connector/package/mongodb_jdbc-v0.0.0.taco
```
terakilobyte: The file is created with this default version now.

bucaojit: The version is now included in the filename.  It depends on the plugin-version that's in `connector/manifest.xml`. 

The default value we have there is 0.0.0, but when it's a tagged build then that value gets set to the actual version ie `1.2.0`.  So there needs to be a little more logic to have the value here be `v0.0.0` by default but if tagged set it to `v$MONGO_TABLEAU_CONNECTOR_VERSION`

Version number when tagged is updated here:
https://github.com/mongodb/mongo-tableau-connector/blob/master/evergreen.yml#L148

bucaojit (changes requested):

terakilobyte:
> The version is now included in the filename. It depends on the plugin-version that's in connector/manifest.xml.

> The default value we have there is 0.0.0, but when it's a tagged build then that value gets set to the actual version ie 1.2.0. > So there needs to be a little more logic to have the value here be v0.0.0 by default but if tagged set it to 
> v$MONGO_TABLEAU_CONNECTOR_VERSION

> Version number when tagged is updated here:
> https://github.com/mongodb/mongo-tableau-connector/blob/master/evergreen.yml#L148

Good catch, I spaced on that. However, I don't think we need to do anything special since we don't really do anything with the filename and end up renaming it for the download center. I updated the script to rename it to the name we use throughout immediately after building the taco file.

bucaojit (approved):
LGTM. 

Yeah your way is simpler for handling the filename change.

PR #58, opened 2024-05-29
SQL-2082: Make UID and PWD optional in connection dialog

bucaojit:
Tested against cluster with OIDC and on remote ADF cluster with username/password.

### DIFF
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -7,7 +7,7 @@
   <field name="authentication" label="Authentication" category="authentication" value-type="selection" default-value="auth-user-pass">
     <selection-group>
       <option value="auth-user-pass" label="Username and Password"/>
-      <option value="auth-none" label="OIDC" />
+      <option value="auth-none" label="Certificate / Token" />
     </selection-group>
   </field>
 
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -4,11 +4,24 @@
 
   <field name="v-mongodb-uri" label="MongoDB URI" value-type="string" category="endpoint" />
 
-  <field name="authentication" label="Authentication" category="authentication" value-type="string" editable="false" default-value="auth-user-pass" />
+  <field name="authentication" label="Authentication" category="authentication" value-type="selection" default-value="auth-user-pass">
+    <selection-group>
+      <option value="auth-user-pass" label="Username and Password"/>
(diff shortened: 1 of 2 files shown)


EthanHardyMongo:
Do you think we need to test with X509 authentication? Or should it work because OIDC which also has no username and password works fine? I have X509 authentication set up locally, so I can try on my end if need be.

bucaojit:
@EthanHardyMongo That would be great if you already have the authentication setup locally.  If you are able to run Tableau, place the taco file that you can [download from here](https://spruce.mongodb.com/task/mongo_tableau_connector_ubuntu2004_sign_patch_0b2ed21975aa346dc4bfd4c252464fd4880dd685_66567d84508d240007bbf9c5_24_05_29_00_57_42/files?execution=0) into `/Users/[your_username]/Documents/My Tableau Repository/Connectors
`.

EthanHardyMongo:
I can't seem to get Tableau to connect via X509 authentication. I'll bring it up in stand up tomorrow.

bucaojit:
I tried adding the keystore properties in `connectionProperties.js`, but the driver keeps logging `authenticationFailed`. 

```
    props["javax.net.ssl.keyStore"] = "C:\\mongodb_pem\\cert.jks"
    props["javax.net.ssl.keyStorePassword"] = "password"
```

At first I thought it was using it's own jks, but I don't see one in the `C:\Program Files\Tableau` directory. 

I did find a way to get it to work in a hacky way which is to specify the JAVA_TOOL_OPTIONS in the command line before running Tableau.  So with this I was able to at least connect:

```
C:\Program Files\Tableau\Tableau 2022.2\bin>set JAVA_TOOL_OPTIONS=-Djavax.net.ssl.keyStore=C:\\mongodb_pem\\cert.jks -Djavax.net.ssl.keyStorePassword=password

C:\Program Files\Tableau\Tableau 2022.2\bin>tableau.exe  -DDisableVerifyConnectorPluginSignature=true
```

I'll create a ticket to investigate getting it to work through the dialog window.  I was thinking of having the selector have an option for `X509` where is will have inputs for `keyStore` and `keyStorePassword`.  Hopefully that way is possible.

terakilobyte (approved):
LGTM, apologies for the delay!

EthanHardyMongo (approved):
LGTM! Thanks for looking into X509!

PR #59, opened 2024-11-12
SQL-2419: Move the logging info UI for the Tableau Connector to "Advanced" tab

bucaojit:
Adds advanced tab that includes the logging options. 
Added `server` to connectionResolver to fix a warning during connector packaging.

### DIFF
--- a/connector/connectionFields.xml
+++ b/connector/connectionFields.xml
@@ -1,29 +1,27 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <connection-fields>
-
+  <!-- Main tab fields -->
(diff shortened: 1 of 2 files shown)


terakilobyte (approved):
LGTM. I actually like the new layout a little more.

I do wonder if Advanced should be before the Initial SQL.

bucaojit:
@terakilobyte I looked through the connector [dialog docs](https://tableau.github.io/connector-plugin-sdk/docs/mcd), unfortunately I don't see a way to set the ordering of the tabs.  I think for now the layout is fine.

EthanHardyMongo (approved):
LGTM!

PR #61, opened 2024-12-10
SQL-2513: Rename v-mongodb-uri to `server` for use in Connections label

bucaojit:
This change will set the server name in the Connection section.  It was initially `v-mongodb-uri` since it contained not only the server name.  But in order to be displayed I'm switching it to the expected `server` it should be enough to help users differentiate their connections.


terakilobyte (approved):
LGTM

EthanHardyMongo (approved):

PR #67, opened 2025-10-28
SQL-2922: Split auth mechanisms into separate connector dialogue

bucaojit:
The dropdown now includes four authentication options:

- Username and Password
- X509: displays password field when checkbox is selected
- GSSAPI: shows username field for principal name input
- Token: no additonal fields 

Manual Testing:
- Tested kerberos with the JDBC changes in [SQL-2920](https://github.com/mongodb/mongo-jdbc-driver/pull/367) with local kerberos on Mac.
- Tested PEM in password field and PEM file in properties file.


connector/connectionFields.xml
```diff
@@ -3,21 +3,39 @@
 <connection-fields>
   <!-- Main tab fields -->
   <field name="server" label="MongoDB URI" value-type="string" category="endpoint" />
-  <field name="authentication" label="Authentication" category="authentication" value-type="selection" default-value="auth-user-pass">
+  <field name="authentication" label="Authentication Mechanism" category="authentication" value-type="selection" default-value="auth-user-pass">
     <selection-group>
       <option value="auth-user-pass" label="Username and Password"/>
-      <option value="auth-none" label="Certificate / Token" />
+      <option value="auth-none" label="X509" />
```
bucaojit, line 9: These are keywords it seems `auth-none` and `auth-user`.  If I use a different name it will error that a `password is missing`.  Ideally I'd name them something more accurate but I added comments to hopefully keep it clear.

connector/connectionFields.xml
```diff
@@ -3,21 +3,39 @@
...
+      <option value="auth-user" label="GSSAPI" />
+      <option value="auth-token" label="Token" />
```
bucaojit: I realized as I was writing the above comment that this would be an issue since it will require a password.  I'll look for a way to fix this tomorrow.

bucaojit:
Here are screenshots of the changes. 

X509_Checked
Token_NoFields
X509_Unchecked
GSSAPI_PrincpalName

connector/connectionFields.xml
nbagnard: Token is for OIDC, right?
Can you use `auth-none` more than once with different labels?
Also, what do you think we change the drop-down menu name to "OIDC", it think that it might be less confusing for users than token now that we have different option for each auth type.

nbagnard (approved):
LGTM!

connector/connectionFields.xml
bucaojit: Yes, OIDC is what's left for `Token` previously it was meant to be for Kerberos and OIDC but now Kerberos is in it's own drop down so I'll rename this to OIDC. 

Using `auth-none` here confuses the interface because it's used later as a conditional for the x509 toggle. 

Luckily, I found one more keyword that looks to do what I want `auth-integrated`.  Hopefully we won't need any more drop-downs here.

EthanHardyMongo (approved):
LGTM!

